PR #1, opened 2019-02-13
Grackle 3.0 - PR

aemerick:
Migrated from Bitbucket:

Pull Request for functional Grackle interface into Enzo-E: 

[Bitbucket PR documentation](https://bitbucket.org/cello-project/cello-src/pull-requests/14/updating-to-grackle-3x-and-improving/diff)


aemerick:
There are a couple of compute pressure functions which are separate from `EnzoComputePressure` in the code, `ComputePressure` and `ComputePressureDualEnergyFormalism`, that I believe are copy-overs of the equivalent bits of code from Enzo. These are both only called in `EnzoMethodComovingExpansion`. Can these be safely switched over to `EnzoComputePressure` ?

brittonsmith:
@aemerick, I think the `ComputePressure` functionality can unified. It looks like `Grid_ComputePressureDualEnergyFormalism` doesn't even exist in enzo-dev anymore.

aemerick:
@brittonsmith Thanks for the look-over. I'll make another pass myself and remove the redundant pressure calculation before calling this good to go. 

One final thing I would like to do for the temperature / pressure / cooling_time fields, however, is to add in a bit of code somewhere to ensure that these are re-computed with the correct values if the user requests them in output. @jobordner Where is the best place to put this check and subsequent call to re-compute these fields? To me it looked like the IO control was handled almost entirely on the Cello side of things, so I'm sure what the best way to do this is.

jobordner:
@aemerick I'm still unsure how best to handle ensuring that derived fields are up-to-date on output.  Creating a "derived" field group would make sense, so the output phase could check if fields are in that group and update them if needed, though there needs to be a mechanism for associating the field with a corresponding "Compute" object (while Methods and Solvers are named, Compute objects currently are not, though it's easy to add that).

aemerick:
Thanks @jobordner I'll look into implementing this using the grouping and by adding names to the compute objects.    Where in the control flow (i.e. which of the IO routines) would it be best to: 1) check if any fields are grouped as "derived", and then 2) call the associated Compute objects?

aemerick:
@brittonsmith In going through and removing the redundant ComputePressure functions that I mentioned earlier, I noticed that the computation of pressure in both Enzo and Enzo-E in their respective ComovingExpansion functions (`enzo_EnzoMethodComovingExpansion` and `Grid_ComovingExpansionTerms.C` respectively) calculates the pressure as time-centered ( 0.5 * (t_new + t_old)).    However, I'm not sure how to do that in the new `enzo_EnzoComputePressure` routine since it is now Grackle-based. Is this an issue?

aemerick:
To answer my own question for how to implement the time-centered computation I could just add an option to have `enzo_EnzoComputePressure` calculate an interpolated pressure at a given time by setting up two `grackle_fields_` structs, one with the current baryon fields and one with the old baryon fields, sending both to Grackle, and then compute the interpolated pressure. 

Before coding this up, however, I just want to double check that this functionality is still needed, or if we'd be O.K. without the time-centered pressure. What do you think @brittonsmith @gregbryan ?

gregbryan:
I think the functionality will be needed for adaptive timesteps, but may not be needed now.

brittonsmith:
@aemerick, I think the method you describe of having two grackle field structs and two calls makes the most sense. I support that.

aemerick:
@brittonsmith @gregbryan. The latest changes include the ability to compute temperature and pressure using fields from any point in the stored history, which brings back the computation of the time-centered pressure that is used in computing the comoving expansion terms.

I've also changed the units behavior in ComputeTemperature to compute the temperature field in Kelvin when used without Grackle, since Grackle's compute_temperature function returns in Kelvin, and since this is the way its done in Enzo.

aemerick:
Update: 

- Recent commits have enabled restarting with Grackle (pending full checks with the restart bug). What this now does is check if the simulation is in a restart, and if so, then rebuilds all of Grackle's internal data structures. This is a MUCH easier way of having to write PUP methods for all of Grackle's structs that just stay internal to Grackle.

- Most recent commit also has enabled rudimentary functionality for having a "derived" fields grouping that is tied to the compute objects by sharing the same name. This allows each block to check if a derived field is being written during output and re-computes this field to be fully updated if so. This may mean some redundant computation of the field, but likely only when output is being done on every or nearly-every timestep.

aemerick:
- compute cooling time compute now working. 

This completes the last to-do (that I can think of) before this can be given a serious look over.

jwise77:
Just a comment: would this PR be a good candidate for rebasing / squashing the commits?

aemerick:
> Just a comment: would this PR be a good candidate for rebasing / squashing the commits?

likely.... maybe I should have paid better attention yesterday. But I'm able to do that if its a good idea.

jobordner:
Regression tests run correctly if the CONFIG_USE_GRACKLE ifdef's are removed from EnzoComputeTemperature and EnzoComputePressure (but leave the ones for the extra grackle parameters).  Testing for (enzo::config()->method_grackle_use_grackle) is sufficient to call the right code.  Fix these and this PR will be good to go.

aemerick:
@jobordner Can you let me know which test(s) were showing the weird values / errors when running with use_grackle = 0? I'm trying to reproduce this locally.

jobordner:
> @jobordner Can you let me know which test(s) were showing the weird values / errors when running with use_grackle = 0? I'm trying to reproduce this locally.

I haven't tried use_grackle = 0 (that was working fine for me before), but parallel runs with use_grackle = 1 can crash in EnzoConfig::pup() when trying to pup the chemistry_data object.  Example test "./build.sh test/test_method_ppm-8.unit".

aemerick:
> > @jobordner Can you let me know which test(s) were showing the weird values / errors when running with use_grackle = 0? I'm trying to reproduce this locally.
> 
> I haven't tried use_grackle = 0 (that was working fine for me before), but parallel runs with use_grackle = 1 can crash in EnzoConfig::pup() when trying to pup the chemistry_data object. Example test "./build.sh test/test_method_ppm-8.unit".

Ah I see. I've found this error and I think I have a fix. Testing now.

aemerick:
@jobordner Ok I tested in parallel when compiled with grackle (double prec only) running both with and without grackle as a method and PUPing seemed to work OK and I did not see any of that weird time value errors from earlier, nor any weird hydro solver errors. I *think* this should function now.

jobordner:
@aemerick Regression tests look fine--huzzah!

aemerick:
@jobordner Yay!! That's good news. 

@bwoshea That was the last to-do needed before a merge

PR #2, opened 2019-02-13
Mercurial to Git updates:

aemerick:
- Changing .hgignore into .gitignore
- Removing mercurial requirements in SConstruct in favor of git,
  and changing commands to grab current changeset

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,3 @@
-# use glob syntax.
-syntax: glob
 Dan_test
 *.a
 *.o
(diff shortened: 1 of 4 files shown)


aemerick:
Thanks for the recommendation @jobordner . I generalized this now

.gitignore
```diff
@@ -0,0 +1,19 @@
+Dan_test
```
brittonsmith: Is this a real thing?

aemerick: Ya... Not sure, it was in the original .hgignore though....

brittonsmith: I grepped around for anything that might make something like this and couldn't find anything. I think you get rid of it.

PR #6, opened 2019-03-04
Style change in Cello block name

aemerick:
Changes style in the instance of running with only a single root-grid block (e.g. `root_block = [1,1,1]`). This is likely never going to happen in any real run since that means running on a single core, but could be possible if doing development tests (as I am doing).

Previously the root grid block name in this case would be   B___, but this changes the name to B0_0_0, which I feel is a bit more consistent with how the naming convention works when there are more than one root blocks, and is more consistent with the refined levels when there is only one root block.

Additionally, this fixes an issue with the Enzo-P frontend in yt, which looks specifically for the B?_?_? naming convention. Sims with `root_block = [1,1,1]` would fail to be recognized because of this. 

In addition, the comments in the yt frontend indicated to me that B___ blocks are assumed to be at a negative refinement level. I'm not too sure of what this actually means on the Enzo-E side of things though, but if this is the case then this also fixes confusion there.

Let me know if this change is worthwhile. Or if there should just be some modifications to the yt front-end to catch this edge case.

### DIFF
--- a/src/Cello/mesh_Index.cpp
+++ b/src/Cello/mesh_Index.cpp
@@ -440,10 +440,14 @@ std::string Index::bit_string(int max_level,int rank, const int nb3[3]) const
 
   for (int axis=0; axis<rank; axis++) {
 
-    for (int i=nb3[axis]-1; i>=0; i--) {
-      int shift = (level >= 0) ? i : i-level;
-      int bit = (a_[axis].array & ( 1 << shift));
-      bits = bits + (bit?"1":"0");
+    if(nb3[axis]-1 < 0){
+      bits = bits + "0";
+    } else{
+      for (int i=nb3[axis]-1; i>=0; i--) {
+        int shift = (level >= 0) ? i : i-level;
+        int bit = (a_[axis].array & ( 1 << shift));
+        bits = bits + (bit?"1":"0");
+      }
     }
 
     for (int i=0; i<max_level; i++) {


jobordner:
Blocks could certainly use an update in naming, though this particular change would actually break slightly with the current pattern.  Blocks are currently named Ba:x\_b:y\_c:z, where (a,b,c) is the index in binary into the array-of-octrees of the root-level octree containing the block, and x, y, and z are bit-strings of the position of the block in its parent, grandparent, etc. back to its root-level octree (":" is only included for refined blocks).  For a [4,4,4] octree array the lowest index root block is B00\_00\_00; for a [2,2,2] array it's B0\_0\_0; so for [1,1,1] it's B\_\_.  Note that levels are relative to the root level blocks, so for the [1,1,1] array, B\_\_ would be level 0, but  would be -1 in a [2,2,2] octree array, and level -2 in a [4,4,4] array, etc.  However, to be consistent with removing ":" for root-level (and coarser) blocks, B\_\_ could perhaps be renamed simply to "B"  (or not to "B"...)

An alternative proposed naming convention I'd like to switch to at some point would be something like Bc, where c is an octal string of (xi,yi,zi) bits.  E.g. the current block name of B101_010_110 (in a [1,1,1] octree array) would be B534  (5=101; 3=011; 4=100).  Then it's more obvious that, e.g., B512346 is a child of B51234, and names would be reduced to about 1/3 the length.  This does not explicitly encode the position in the root grid octree array, but would not really need to be, provided the octree array dimensions are available as meta-data.

aemerick:
Thanks for the explanation @jobordner .. I definitely did not have a complete grasp of the naming scheme prior to making this change... Given this, I'm happy rejecting this PR in favor of the current scheme and instead submitting a PR to yt to better handle the case of a single [1,1,1] root block.

PR #9, opened 2019-03-29
VL+CT MHD Integrator

mabruzzo:
This PR for adds the implementation of the VL + CT (van Leer + Constrained Transport) MHD integrator detailed in Stone & Gardiner (2009). 

**Brief summary of the code included in this pull request:**

1. The VL+CT method has been implemented in the `EnzoMethodMHDVlct` class.
2. The `CelloArray` class has been defined to encapsulate the array class and is used to write the bulk of operations in this method more clearly and concisely (and can either wrap around previously allocated data or own its own data). 
    - The class is implemented as part of a new Cello "component" (the "array" component) and it is now stored in a file with the "array" prefix. *(It may make more sense to place it in an existing component - if that is the case, let me know and I'll move it)*
    - Unit tests are provided in `src/Cello/test_CelloArray.cpp`. Although they don't provide complete coverage over all use-cases, they are extensive enough to be useful for preventing regressions in the design. Extending coverage would probably be useful in the future. These unit tests have been correctly added to `src/Cello/SConscript` and can be built by invoking `./build bin/test_CelloArray` from the root directory (like all existing unit tests). They have also been added to `test/SConscript` so that the unit test is run with all other unit tests when `make test` is invoked. 
3. This method makes use of decoupled implementations of Riemann Solvers, reconstruction algorithms, and equations of states that are subclassed from the abstract base classes `EnzoRiemann`, `EnzoReconstructor`, and `EnzoEquationOfState`. These classes form the basis of a loose framework that can easily (and selectively) be reused (and extended) in other hydrodynamic/MHD methods
    - Specifically, I provide implementations of the HLLE, HLLC, and HLLD approximate Riemann Solvers (`EnzoRiemannHLLE`, `EnzoRiemannHLLC`, `EnzoRiemannHLLD`), nearest neighbor and PLM reconstruction algorithms (`EnzoReconstructorNN`, `EnzoReconstructorPLM`), and an equation of state for an ideal gas ( `EnzoEOSIdeal`)
    - `EnzoEquationOfState` sub classes are also responsible for applying the dual-energy formalism, if applicable ( `EnzoEOSIdeal` does in fact support this)
    - The act of updating quantities from the flux divergence has also been factored out into it's class called `EnzoIntegrableUpdate` to allow for reuse in other methods.
    - Note that none of these classes are derived from the existing Cello framework (solver classes, compute classes, or method classes) as their interfaces are not ideal for these operations. However, all of these classes have `PUP` methods.
4. The constrained transport functions are housed in the `EnzoConstrainedTransport` class.
5. Several new initial condition classes are provided:
    - The `EnzoInitialInclinedWave` class initializes inclined linear HD/MHD waves and inclined circularly polarized Alfven waves, which are used for debugging and convergence tests.
    - The `EnzoInitialShockTube` class initializes axis-aligned shock tube test problems. Currently it initializes the Sod shock tube problem and the shock tube shown in figure 2a of Ryu & Jones (1995). 
    - The `EnzoInitialBCenter` class sets up the cell-centered magnetic fields from previously initialized face-centered magnetic fields OR sets up both the cell-centered and face-centered magnetic fields from expressions specified in the configuration file for components of the magnetic vector potential. It also provides the ability to update the values of a partially initialized `"total_energy"` field with the values of the initialized magnetic energy. (This convenience is provided since the cell-centered magnetic field is an average of the face-centered magnetic energy, and therefore may be non-trivial to compute)
    - The `EnzoInitialCloud` class sets up a 3D spherical cloud embedded in a hydrodynamic/MHD wind.
6. `EnzoCenteredFieldRegistry` is a utility class that holds functions that assist with the initialization of the  `EnzoReconstructor` and `EnzoRiemann` subclasses. It houses a list of cell-centered quantities that should be updated as new fields are added that are relevant to either of those classes (e.g. internal energy or cosmic ray energy) get added. It also provides methods useful for initializing `EnzoMethodMHDVlct`.
7. `EnzoPermutedCoordinates` and `EnzoFieldArrayFactory` are utility classes. The former holds functions that assist with writing directional operations mesh-based operations (e.g. reconstruction, flux calculations), that are generalized with respect to direction. The latter houses methods that simplify the construction of  `CelloArray` instances that wrap fields.
8. Passively advected scalars are supported. However, there is not yet support for making sure that collections of passive scalars maintain a constant normalization (this is left to a future pull request).

**Documentation update**

  - The user documentation has been updated to reflect the introduction of the `EnzoMethodMHDVlct` method (referred to in the parameter file as "mhd_vlct") and the the new initial conditions
  - The reference guide on the input parameters has been updated with descriptions of the input parameters of the method and each of the new initial conditions.
  - A page has been added to the developer's guide describing how to use ``CelloArray`` (this was adapted from the proposal circulated on slack).
  - A page has been added to the developer's guide describing the hydrodynamic/mhd infrastructure introduced in this pull request (and that can be optionally/selectively reused in other methods). The guide provides a general overview of the infrastructure, defines some terms used throughout the implementation (these terms are also defined in the implementation), explains the rationale behind some of the more significant design decisions, describes the interfaces of each class in the framework, and describes how to extend the relevant classes in the framework (e.g. adding a new Riemann Solver). *(If nobody thinks this is useful, I'm happy to remove it).*

**New Tests**
6 sets of repeatable tests have been included in the `input/vlct` sub-directory. These tests exist as standalone python scripts print details about whether individual tests have passed and that exit with error codes signaling whether or not the full set of tests have passed. The scripts are:

  1. `run_MHD_linear_wave_test.py` - this checks the L1 Error norms of several MHD waves after propagating at an oblique angle to the simulation domain. Note that one of these tests fails. In that test, the L1 error norms of an inclined left and right propagating fast magnetosonic wave are compared - the norms of the L1 error vector should be identical, but there is a slight discrepancy (the values are 3.30e-8 & 3.28e-8). While this should be addressed (it probably points to a minor issue in the constrained transport), we don't think this should stop the acceptance of this pull request because all of the other tests pass. When this test fails currently fails, the user is notified of the failure and the fact that the test has never passed; the return code of this script is not affected by failure of this test. Once the pull request is accepted, I plan to open an error about this topic.
  2. `run_MHD_shock_tube_test.py` - this checks the L1 Error norms of the axis-aligned MHD shocktube test problem taken from figure 2a of Ryu & Jones (1995). The test is repeated for the x, y, and z axes.
  3. `run_passive_advect_sound_test.py` - this checks the norm of the L1 Error of a passively advected scalar after it has been advected by an axis-aligned sound wave for a full wavelength. The test is repeated for the x, y, and z axes.
  4.  `run_HD_linear_wave_test.py` this checks the L1 Error norms of several HD waves after propagating at an oblique angle to the simulation domain. Magnetic fields are uniformly set to zero while this runs
  5. `run_dual_energy_shock_test.py` Runs several variants of the sod shock-tube test in a supersonic reference frame to test the implementation of the dual energy formalism.
  6. `run_dual_energy_cloud_test.py` Runs a cloud-crushing problem (a cloud embedded in a hot wind) for ~30 cycles (technically it runs for a sixteenth of a "cloud crushing time") and checks the asymmetries that develop. This turns out to be a useful test of the dual energy formalism.

These tests don't lend themselves to being implemented in the test suite in its current state (as the test suite currently just uses the built-in functionality of enzo-e to check the final times/cycles of a simulation). Including these tests in the existing current test suite problem may require some rewriting of how tests are counted. To help make the tests easier to adapt to the final testing framework, the majority of them employ a command line program, `tools/l1_error_norm.py` to compute norms of the L1 Error. All of these python scripts are written with python3 because (at least at one point) bugs were present in `yt`'s enzo-p frontend when using python2 that were not present when using python3. Unfortunately, this introduces some additional incompatibility with the existing test suite since the employed version of SCons is only compatible with python2.

A self-contained temporary solution has been provided to make sure that these tests are run even though they are not yet formally part of the testing framework. This solution consists of:
- introducing a simple bash script called `test/run_vlct_test.sh` to run each of the above tests. For each of the tests, the script pipes the stdout and stderr into a temporary file, named `test/<test-name>.test_log`, and prints whether the test has passed/failed. After running all of the test, the script indicates prints whether any have failed and reflects this in its exit code.
- modifying `build.sh` to call this script when `./build.sh test` is executed and `"$CELLO_PREC" = "double"`. It is called after the existing test suite have finished running (and after the results for those tests are printed). If any of the vlct tests failed, the exit code of `build.sh` is set to 1.
- adding a line to `build.sh` to delete any temporary files that hold the test outputs when `./build.sh clean` is executed.
- adding a few minor tweaks to `.circleci/config.yml` to setup the necessary python environment for these tests.  I encountered some difficulties with getting parallel simulations to run, an environment variable is also set to skip `run_dual_energy_shock_test.py` (it was necessary to run this test in parallel so that it completed in a reasonable amount of time). I don't think we are losing much coverage by skipping this test.

The file names used to store the test outputs and the way that these test results are counted intentionally differs from the existing framework so that I don't break the existing machinery. **Note:** If the way these tests have been added to the continuous integration is going to block the approval of this PR, my preference would be to move this part of the request to a separate PR.

Additionally, an input file is provided, in `input/vlct/orszag-tang`, to run the Orszag-Tang Vortex test problem. There is no automated check that the results are correct (that's probably something to be added once an analogue to Enzo's answer testing framework is provided). For convenience, a script is provided that produces two plots from the simulation results that can be compared with two figures in Stone et al (2008).

**Dependence on other pull requests**
This pull request has included bugfixes included in pull request #19 

**Future Work**
I have begun a separate branch to make additional changes related to this integrators beyond the scope of this pull request (the changes will be submitted in a future pull request). These changes include:
 - Support for Grackle (*It is almost supported in cases with fixed adiabatic index. The only thing stopping it is a refresh bug. When all fields are added to the refresh list, cello will try to refresh temporary fields that have already been deallocated. An issue will be filed once the PR is accepted*).
  - Add support for running the solver as a purely hydrodynamic solver. I have a separate branch where this is nearly done (lots of documentation changes still need to be made).
 - Transition from grouping hydrodynamic quantities in ``Grouping``s of field names to organizing them in maps of ``EnzoArrays``. This transition will make the control flow easier to follow and should help any future efforts to use GPUs (by allowing better control over where data is stored).
  - Potentially compile all of the Riemann Solvers in a separate static library and link the rest of the enzo object files to that library (analogous to Cello). This would allow us to take the headers defining the templates out of the main `_enzo.hpp` header file, and would therefore allow us to avoid recompiling the entire enzo layer every time a Riemann Solver is tweaked. It also wouldn't carry any runtime penalty since they are all subclasses of the `EnzoRiemann` class, anyway.


forrestglines:
@mabruzzo General question about the `EnzoArray`, slicing, and general implementation. Are there any transposes or large striding across dimensions? These might introduce large performance hits.

Specifically I'm looking through the reconstruction step, I see that the stencil is across the x dimension. How does the reconstruction in y and z work?

forrestglines:
Ok, I see what that's doing now. Good!

Another question about how accessing a slice with the () operator: Is it just some extra inlined integer arithmetic, like in Athena++? Are there any extra pointer dereferences going on?

mabruzzo:
Accessing an element of a slice with the () operator, just performs a single pointer dereference with some extra inlined math. It's exactly like accessing an element in a normal array.

Below is a benchmark I recently ran for EnzoArray's () operation.
image

In the plot I compare how long it takes to iterate over an `(N,N,N)`. I examined cases of a pointer, `Boost::multi_array`'s [ ][ ][ ] operator, `Boost::multi_array`'s () operator, `EnzoArray`'s () operator, and `EnzoArray_neg`'s () operator. (The last case is just `EnzoArray` with negative indexing). 

The top panel, shows how long it takes for the operation to occur as a function of `N`. The plotted point is the minimum time it took out of 100 trials. The bottom panel shows the ratio of how long each trial takes compared to the pointer. (Note I don't really believe that the multi_array is significantly slower for small `N`.

The script to replicate this can be found [here](https://github.com/mabruzzo/array_benchmark). 

I don't actually buy that the boost:multi_array is 5% slower for

mabruzzo:
This pull request is ready for review

mabruzzo:
The most recent series of commits added changes that I had accrued over the last several months including. The 2 largest changes are adding support for the dual-energy formalism and overhauling the Riemann Solver (the machinery is simpler and ~30% faster). The website docs have been updated accordingly. This is ready for review.

I will try not to introduce any more changes to the code.

src/Enzo/enzo-p.cpp
```diff
@@ -37,6 +37,13 @@
 
 #include "charm_enzo.hpp"
 
+// The following needs to be included once and only once
+// This may not be the perfect place for this, but it is when it is included in
+// multiple object files
+#define CK_TEMPLATES_ONLY
+#include "enzo.def.h"
+#undef CK_TEMPLATES_ONLY
+
```
jobordner, line 46: This include should be added to `test_EnzoUnits.cpp` as well.

mabruzzo:
The GitHub interface appears to have some difficulty with distinguishing the commits from the Solver-DD PR and commits uniquely from this PR.

Following my conversation with Forrest and Phillip, I'm going to try closing this PR and resubmit it as a new PR.

PR #14, opened 2019-05-08
Run test suite on circleci.

brittonsmith:
This PR runs `make test` and counts the number of failures and incompletes to determine pass or fail. I've set the maximum number of failures + incompletes to be 11 for the known failing tests. We can lower that when they get fixed. This will also now test both single and double precision builds.


brittonsmith:
Here's the tail of the `make test` output:
```
14:32:05 linux_gnu double       FAIL:  5      test/fail.linux_gnu-double Incomplete: 6 test/incomplete.linux_gnu-double Pass: 40246 test/pass.linux_gnu-double   
   CRASH: test/test_initial_music-111.unit
   CRASH: test/test_initial_music-112.unit
   CRASH: test/test_initial_music-114.unit
   CRASH: test/test_initial_music-121.unit
   CRASH: test/test_initial_music-141.unit
   CRASH: test/test_initial_music-211.unit
   CRASH: test/test_initial_music-222.unit
   CRASH: test/test_initial_music-411.unit
   CRASH: test/test_method_cosmology-1.unit
END   Enzo-P/Cello ./build.sh: arch = linux_gnu  prec = double  target = test time = 7.80 min
```

Is there something else that needs to be done to get the music and cosmology tests to run?

jwise77:
Cosmology runs only work in single precision. I think this is why those tests are failing.

brittonsmith:
@jwise, this is functional now if you want to have a look.

bwoshea:
This looks good to me.  Any objections to me merging?

bwoshea:
@jobordner , I'm trying to translate your comment to action.  Do you think we should change something in the current PR, or is that a cautionary note for the future?

jobordner:
@brittonsmith It looks like circleci parallel runs aren't running correctly, failing at startup, though they work fine on my server ( [Enzo-E / Cello Test Results: Open pull requests](http://client65-90.sdsc.edu/~bordner/enzo-e-test/Dir.PR/test.html) ) Is there a way to view the output of individual tests?  How was charm++ built, I suspect that may be the problem.

brittonsmith:
@jobordner, if you'd like to see the charm++ build in the log output, just comment out the `restore_cache` and `save_cache` commands that surround `install-dependencies` in the config.yml file.

If you want to see the full output for the test-suite, you can click on the "details" links next to the circleci jobs on this page. There, scroll down to the last block and at the bottom there will be a link to download the full text of the log output. [Here](https://circleci.com/api/v1.1/project/github/enzo-project/enzo-e/101/output/107/0?file=true) is the log for the last run of test-suite-1.

jobordner:
@brittonsmith Sorry, I meant viewing the output of an individual test, e.g. the contents of, say, the `test/test_method_ppm-8.unit` file.  Maybe the only way is to temporarily update the config file to write the output of a parallel run to stdout.

brittonsmith:
@jobordner, ah my mistake. I think the best way to do this would be to rerun the test-suite with ssh enabled, so you can login to the node it ran on once the job has finished. If you want to do this, go to the page of any given job, and in the upper-right corner there is a dropdown menu next to "Rerun workflow". You'll find the option there and it will provide you with ip/port to login at the end of the log output.

jobordner:
@brittonsmith Exactly what I was looking for, thanks.  I tried the ssh thing, and a basic PPM parallel run with Charm++ failed with the following:

> ../local/charm-6.9.0/bin/charmrun +p4 bin/enzo-p input/method_ppm-8.in 
Charmrun> scalable start enabled. 
ssh: connect to host localhost port 22: Cannot assign requested address
Charmrun> Error 255 returned from remote shell (localhost:0)
Charmrun> Reconnection attempt 1 of 3

The problem is Charm++ parallel runs needs to ssh to localhost, but circleci didn't allow that.   Though I figure it's possible to configure circleci for this, given lines like `SSH_TARGET: "localhost"` in the documentation.

brittonsmith:
@jobordner, I got the parallel run you posted above to run on the circleci machine by adding the `++local` flag. I wasn't totally clear on how to integrate that into the testing. If that works, can you set that up?

brittonsmith:
If I read this right, @jobordner's last changes now have the tests running as expected, that is, the single precision test suite runs with no failures and the double precision failures are just the known issue with reading MUSIC ICs. What is the best course of action now? Should we add an allowance for the known music failures in double precision and then merge this?

gregbryan:
That plan seems sensible to me.

jobordner:
@brittonsmith I'm working on a fix for the HDF5 precision issue, which I expect to be done later today.  I can either push to this branch or create a separate PR.

brittonsmith:
@jobordner, I think it's fine to push that here so that we can merge this with all the tests passing.

brittonsmith:
@bwoshea, James' last change has made this a fully working test suite in both single and double precision. I think we should merge this as soon as possible so we can starting running this with the other PRs.

Issue #15, opened 2019-05-08
Passing particles from parent to child during ICs


aemerick:
Currently, there is a bug that prevents passing particles from a parent blocks (i.e. the root grid) to child blocks during initialization. This is preventing ICs with both a large number of particles and a large number of highly refined blocks, but a reasonable number of root-grid blocks. Although reading these particles into each child block works, this can take an incredibly long time for large particle lists / block counts. 

More generally there may be a need to do something creative to read in VERY large particle ICs for VERY large root-grids, but that is longer term and solutions may be problem-type specific.

jwise77:
I _think_ this was fixed in #49. Tentatively closing.

PR #18, opened 2019-05-12
Particle ICs Bugfix

aemerick:
Fix for Issue #15. Particles initialized on the root grid should now be properly passed down to child blocks during initialization.

This has been tested on a simple problem with two particle types on a grid with nested ("mask") refinement to level 1.


src/Cello/mesh_Block.cpp
```diff
@@ -102,6 +102,10 @@ Block::Block ( MsgRefine * msg )
 
   if (is_first_cycle) {
     apply_initial_();
+
+    if (this->level() > 0){
+      msg->update(data());
+    }
```
aemerick: In my test I only saw calls to p_set_msg_refine. The particle ICs worked without the inclusion of this if statement in Block::Block. However, the similarity of these two functions made me think it should go here as well.

bwoshea:
@gregbryan and @jwise77 I've added you as reviewers on this.  Please take a look!  I am also a reviewer.

bwoshea:
This looks good to me!

aemerick:
@bwoshea I think this is good to go!

aemerick:
@jobordner It looks like this is failing in the `test_particle-amr` tests and I suspect it has something to do with the fact that in these tests, particles are created on each level, level-by-level, so there is no need to pass information from particles at levels above into more refined levels, and apparently doing so causes problems.. 

I'm not too familiar with these tests though or the `Trace` method, so I'm not sure if there could be a simple fix by tweaking something in the parameter file(s) that would work. Otherwise this might need some parameter that describes how particles are initialed in a given initialization method (root grid and passed, or level-by-level), and do the `msg->update(data())` calls accordingly depending on this  parameter.

jwise77:
Should we wait to merge until the `test_particle-amr` tests succeed? Or should we go ahead and merge?

aemerick:
Just an update. It looks like this is still failing in the particle AMR static and dynamic tests, but as before its not clear to me why

bwoshea:
> Just an update. It looks like this is still failing in the particle AMR static and dynamic tests, but as before its not clear to me why

@jobordner would you mind taking a look at this PR?

src/Cello/mesh_Block.cpp
bwoshea: @jobordner would you mind commenting on this?

aemerick:
@jobordner @bwoshea:   After pulling in the latest master it looks like whatever issue with the testing this was having is now resolved. Not sure what did it, but I guess this is good to go (and means that there shouldn't be the same issue in PR 49 anymore)

jobordner (approved):

bwoshea:
Note: as per my discussion with @aemerick , I believe we are going to scrap this PR because its content is included in PR #49 .  However, since that is still pending we are going to defer merging or declining this PR for the time being.  @aemerick can you confirm?

jwise77:
I'm closing this PR since it has been included with #49.

PR #19, opened 2019-05-17
Face Boundary io

mabruzzo:
This pull request fixes the outstanding problems in Issue #7 (PR #8 had previously addressed some of the problems). Specifically, this pull request:

1. adds a `field_cell_faces` method to the `Data` class to compute the positions of cell-faces
2. refactors `InitialValue` to make use of the `Value` class
3. fixes the bug with initializing face-centered fields for the "value" initial conditions
4. fixes the bugs with refreshing face-centered fields using the "inflow", "outflow", and "reflecting" boundary conditions

Note I have been unable to figure out how to run the test suite successfully. Each time I have tried, I have encountered problems (which are unrelated to which branch I use). 

I have created some tests to demonstrate that these changes work. I wasn't sure they were worth adding them to the test suite (or how to add them to the test suite). The tests can be found here: 
[boundary_test.tar.gz](https://github.com/enzo-project/enzo-e/files/3193254/boundary_test.tar.gz)

To run the tests, modify the included Makefile with the path to your enzo-e executable. These tests illustrate test in that face-centered fields can be successfully initialized by value and that they can be updated by the various boundary conditions. I've also included a test illustrating that the explicit issue cited in #7 for periodic boundary conditions was in fact solved in the prior pull request.

### DIFF
--- a/src/Cello/problem_InitialValue.cpp
+++ b/src/Cello/problem_InitialValue.cpp
@@ -15,82 +15,12 @@ InitialValue::InitialValue
 (Parameters * parameters,
  int cycle, double time) throw ()
   : Initial (cycle, time),
-    parameters_(parameters),
-    num_fields_(cello::field_descr()->field_count()),
-    num_masks_(NULL),
-    mask_(NULL),
-    nx_(0),
-    ny_(0)
+    parameters_(parameters)
 {
   parameters_->group_set(0,"Initial");
   parameters_->group_set(1,"value");
-
-  mask_      = new bool ** [num_fields_];
-  nx_        = new int * [num_fields_];
-  ny_        = new int * [num_fields_];
-  num_masks_ = new int [num_fields_];
-  for (int i=0; i<num_fields_; i++) {
-    mask_[i] = 0;
-    nx_[i] = 0;
-    ny_[i] = 0;
-    num_masks_[i] = 0;
-  }
-  
-  // Initialize fields
-
-  for (int index_field = 0; index_field < num_fields_; index_field++) {
-
-    std::string field_name = cello::field_descr()->field_name(index_field);
-
-    //    parameters_->group_set(1,field_name);
-
-    if (parameters_->type(field_name) == parameter_list) {
-      int num_values = parameters_->list_length(field_name);
-      if (num_values > 1) {
-	num_masks_[index_field] =  num_values / 2;
-	mask_[index_field] = new bool * [num_masks_[index_field]];
-	nx_[index_field]   = new int [num_masks_[index_field]];
-	ny_[index_field]   = new int [num_masks_[index_field]];
-	// loop through value masks
-	for (int index_mask=0; 
-	     index_mask < num_masks_[index_field];
-	     index_mask++) {
-	  int index_value = index_mask*2+1;
-	  if (parameters_->list_type(index_value,field_name) == parameter_string) {
-	    std::string file 
-	      = parameters_->list_value_string(index_value,field_name,"default");
-	    create_mask_png_ (&mask_[index_field][index_mask],
-			      &nx_[index_field][index_mask],
-			      &ny_[index_field][index_mask],
-			      file);
-			   
-	  
-	  }
-	}
-      }
-    }
-  }
 }
 
-//----------------------------------------------------------------------
-
-InitialValue::~InitialValue() throw()
-{
-  for (int index_field = 0; index_field<num_fields_; index_field++) {
-    for (int index_mask = 0; index_mask<num_masks_[index_field]; index_mask++) {
-      if (mask_[index_field][index_mask]) {
-	delete [] mask_[index_field][index_mask];
-	mask_[index_field][index_mask] = NULL;
-      }
-    }
-    delete [] mask_[index_field]; mask_[index_field] = NULL;
-    delete [] nx_[index_field];   nx_[index_field]   = NULL;
-    delete [] ny_[index_field];   ny_[index_field]   = NULL;
-  }
-  delete [] mask_; mask_ = NULL;
-  delete [] nx_;   nx_   = NULL;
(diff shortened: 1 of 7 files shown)


mabruzzo:
@jobordner I've got a question related to resolving the last 2 failing tests of this pull request. This failure can be traced back to subtle differences in the treatment of image masks by `MaskPng` and the earlier version of `InitialValue`. The main question boils down to: **should the specified image mask only extend over the active region or should it also include the ghost zones AND what should the threshold value for masking be?**

To provide some context, I wanted to go over the difference between `MaskPng` and the earlier version of `InitialValue`. To do this, it is helpful to define some terms. In each case, the value of the mask at a position of `(x_val,y_val)` is determined by the value of the pixel closest to `(x_val,y_val)` in an image with shape `(nx_png,ny_png)`. In each case the widths of the image, `(image_width_x,image_width_y)`, and the lower edges `(lower_x,lower_y)` of the image are based on the simulation domain. The mask value is determined by the image pixel with x-index given by: `floor(nx_png * (x_val - lower_x) / image_width_x)` (the y-index is determined similarly).

It seems to me that the choice of `(image_width_x,image_width_y)` and `(lower_x,lower_y)` are different for both cases. Additionally the threshold for masking an image is different. For `MaskPng`, these values correspond to the dimensions of the active zone. Additionally regions are masked if any pixel has a non-zero value.


Judging from the fact that the width of the ghost zones are included in the calculation of `(image_width_x,image_width_y)` in the earlier version of `InitialValue` (see [lines 532-533](https://github.com/enzo-project/enzo-e/blob/master/src/Cello/problem_InitialValue.cpp#L532) ), I believe `InitialValue` was meant to include the ghost zone. However, I'm pretty sure that the calculations of the values of `(lower_x,lower_y)` are inconsistent (see [lines 538 - 539](https://github.com/enzo-project/enzo-e/blob/master/src/Cello/problem_InitialValue.cpp#L538) ). I think I think they actually read:
`offset_b[0] = lower_b[0]-lower_h[0]-gx*hb[0];`
`offset_b[1] = lower_b[1]-lower_h[1]-gy*hb[1];`
Additionally, a threshold value of 65536*3/2 is used to determine if an image should be masked.

I think it probably makes more sense to adopt the convention of `MaskPng` (and it would be easier), but I wanted ask before I change the test results.

jobordner:
Hi @mabruzzo I agree, the conventions of `MaskPng` should be preferred to that of the older `InitialValue`.   I haven't reviewed the mask code in detail and it's been awhile, but I expect a lot of common code could be factored out so that evaluating the mask is only ever done in `MaskPng`.  Thanks for tracking this down!

mabruzzo:
I've modified the expected test result to follow the convention of the MaskPng class - the mask only extends over the active zone and any non-zero pixels indicate that values should be masked. You can replicate this result in the master branch by dropping in the linked, modified version of [problem_InitialValue.cpp](https://pastebin.com/8Ck04Tr5).

At this point, the pull request is complete and is ready to be reviewed.

src/Cello/problem_InitialValue.cpp
```diff
@@ -157,48 +91,47 @@ void InitialValue::enforce_block ( Block * block,
...
+      field.centering(index_field, &cx,&cy,&cz);
 
-      ASSERT1("InitialValue::enforce_block",
-	     "Length of list parameter Initial:%s:value must be odd",
-	     field_name.c_str(),
-	     (list_length % 2) == 1);
+      if (val_array == NULL){
+	xc = new double [nx+2*gx];
+	yc = new double [ny+2*gy];
+	zc = new double [nz+2*gx];
```
jobordner: the gx in this line should be gz

src/Cello/problem_InitialValue.cpp
```diff
@@ -157,48 +91,47 @@ void InitialValue::enforce_block ( Block * block,
 
     } else if (parameter_type == parameter_list) {
 
-      // Check parameter length
+      Value value(parameters_, field_name);
```
jobordner: The `MaskPng` objects are re-initialized each time the `Value` constructor is called, which here is once per block.  For a problem involving lots of blocks (e.g. the load balancing test problem `input/load-balance.in`, which generates over 5000 blocks at initialization) it's re-reading the PNG file lots of times, leading to a performance issue. `InitialValue` originally initialized the masks at the start and saved them; you may want to try something similar with the `Value` objects.

bwoshea:
@jobordner you self-assigned yourself to work on this PR.  I have done so too.  @gregbryan  , I hope you can work on this one as well!

bwoshea:
@jobordner could you take a look and see if Matthew addressed your concern?

src/Cello/problem_InitialValue.cpp
```diff
@@ -157,48 +91,47 @@ void InitialValue::enforce_block ( Block * block,
...
-      // Allocate arrays if needed
-      if (value == NULL) {
-	allocate_xyzt_(block,index_field,
-		       field_data,
-		       &nx,&ny,&nz,
-		       &value, &vdeflt,
-		       &mask,  &rdeflt,
-		       &x,&y,&z,&t);
+	// set val_array as array of 0s big enough for corner centered fields
+	val_array = new double[(nx+2*gx+1)*(ny+2*gy+1)*(nz+2*gy+1)]();
```
jobordner: The second gy should be a gz :-)

jobordner:
Restart tests crash with a seg fault, though I don't know why.  The P=1 one can be run using `./build.sh test/test_restart_ppm-1.unit`.

src/Cello/problem_InitialValue.cpp
mabruzzo: Thanks for catching this

mabruzzo:
I believe that these commits should address each of your concerns.

PR #22, opened 2019-05-23
Grackle bugfix

aemerick:
Fixing bug that cause simulation to hang using Grackle method in AMR runs. compute_done now properly called.


bwoshea:
@fearmayo you are a reviewer on this, having self-assigned it.  I will also be a reviewer.  @brittonsmith , I hope you can do so as well!

bwoshea:
3 approves; merging!

PR #24, opened 2019-06-04
Removing old PPM routines

aemerick:
This PR removes the need for the NEW_PPM ifdef by removing all of the "old" PPM routines in favor of those pulled from the more recent enzo-dev versions. This deletes all of the older fortran routines in favor of the new ones (those that had "_dev" in their name). These same files are renamed to strip the "_dev", removing.

The diff on this is quite large, but this is due to the file renaming.

This *should* not break any tests as I believe everything was compiled with NEW_PPM turned on.


aemerick:
I decided to change all instances of "colour" to "color" for consistency (since I saw both in difference places). I don't really care which spelling we use, but opted for "color".

bwoshea:
@jwise77 , @fearmayo  - you had volunteered yourself to work on this PR.  @gregbryan , I'm requesting that you look at this PR.  Please go through it and take a look.  Thank you!

aemerick:
@fearmayo I did not explicitly check the answers myself, but presumably this is checked with the test-suite? 

Otherwise I think this is good to go on my end.

bwoshea:
Three approves, I'm going to merge this one.

Issue #28, opened 2019-11-18
Scons TypeError


mabruzzo:
While building enzo-e on stampede2, I encountered a python error related to scons at the end of the build process. Everything proceeds as normal until the very end of the build process. Right after the newly-built executable is copied to `bin/enzo-p`, the traceback for a python `TypeError` is output. However then the build script indicates that the installation succeeds. 

This occurs for both choices of compilers.  As far as I can tell, this doesn't actually cause any issues and the executable works as it is supposed to. However, I wanted to document this issue. 

An excerpt of the output illustrating this issue is provided below. The full log from the same build process can be found [here](https://pastebin.com/iaY0hqaf).

> Install file: "build-stampede_build_update/Cello/libcello.a" as "lib/libcello.a"
> /home1/05274/tg845732/charm/bin/charmc -language charm++ -o build-stampede_build_update/Enzo/enzo-p -Ofast -g -Wall -rdynamic -module CommonLBs build-stampede_build_update/Enzo/enzo-p.o build-stampede_build_update/Cello/main_enzo.o -Llib -L/opt/apps/intel18/hdf5/1.10.4/x86_64/lib -L/opt/apps/intel18/boost/1.68/lib -L/home1/05274/tg845732/local/lib -L/usr/lib64/lib -lenzo -lcharm -lsimulation -ldata -lproblem -lcompute -lcontrol -lmesh -lio -ldisk -lmemory -lparameters -lerror -lmonitor -lparallel -lperformance -ltest -lcello -lexternal -lhdf5 -lz -ldl -lpng -lirc -limf -lifcore -lifport -lstdc++ -lintlc -lsvml -lboost_filesystem -lboost_system -lgrackle
> Install file: "build-stampede_build_update/Enzo/enzo-p" as "bin/enzo-p"
> TypeError: signal handler must be signal.SIG_IGN, signal.SIG_DFL, or a callable object:
>   File "/home1/05274/tg845732/enzo-e/scons-local-2.2.0/SCons/Script/Main.py", line 1343:
>     _exec_main(parser, values)
>   File "/home1/05274/tg845732/enzo-e/scons-local-2.2.0/SCons/Script/Main.py", line 1307:
>     _main(parser)
>   File "/home1/05274/tg845732/enzo-e/scons-local-2.2.0/SCons/Script/Main.py", line 1071:
>     nodes = _build_targets(fs, options, targets, target_top)
[...]

PR #29, opened 2019-12-05
Updated Stampede Build Files

mabruzzo:
I updated the configuration files (`stampede_gnu.py` and `stampede_intel.py`) for building Enzo-E on Stampede2 and I have confirmed that Enzo-E is successfully compiled using both compilers. Additionally, I factored out the duplicate code used to search for the ``charm_path`` and ``grackle_path``, from the ``linux_gnu.py`` and ``linux_intel.py`` configuration files into dedicated functions in ``_common_search_paths.py`` (these functions are now used in all four of these configuration files).

I just wanted to note that when building on stampede2, using either compiler, Scons raises an error (detailed in Issue #28 ). This error arises at the very end of the installation and doesn't seem to have any actual effect.

### DIFF
--- a/config/_common_search_paths.py
+++ b/config/_common_search_paths.py
@@ -1,28 +1,26 @@
 import os
 
+def _check_dirs(search_paths):
+        for search_path in search_paths:
+                if os.path.isdir(search_path):
+                        return path
+        return None
+
 def charm_path_search(home):
 	# function that searches for charm_path in locations common across
 	# multiple architectures
 	charm_path = os.getenv('CHARM_HOME')
 
 	if (charm_path is None) and (home is not None):
-		if os.path.isdir(home + '/Charm/charm'):
-			charm_path = home + '/Charm/charm'
-		elif os.path.isdir(home + '/charm'):
-			charm_path = home + '/charm'
-		elif os.path.isdir(home + '/local/charm'):
-			charm_path = home + '/local/charm'
-		elif os.path.isdir(home + '/src/charm'):
-			charm_path = home + '/src/charm'
-		elif os.path.isdir(home + '/source/charm'):
-			charm_path = home + '/source/charm'
+                temp = ['Charm/charm', 'charm', 'local/charm', 'src/charm',
+                        'source/charm']
+                paths = [os.path.join(home, elem) for elem in temp]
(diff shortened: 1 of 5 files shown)


bwoshea:
@aemerick and @jwise77 , could you please verify you can compile on Stampede2 and, if so, approve this?  I will do so as well.

config/_common_search_paths.py
```diff
@@ -0,0 +1,50 @@
+import os
+
+def charm_path_search(home):
+	# function that searches for charm_path in locations common across
+	# multiple architectures
+	charm_path = os.getenv('CHARM_HOME')
+
+	if (charm_path is None) and (home is not None):
+		if os.path.isdir(home + '/Charm/charm'):
+			charm_path = home + '/Charm/charm'
```
brittonsmith, line 10: I realize this is mostly moved code and not new, but a safer way to do this might be `os.path.join(home, <dir>)`.

config/_common_search_paths.py
```diff
@@ -0,0 +1,50 @@
+import os
+
+def charm_path_search(home):
+	# function that searches for charm_path in locations common across
+	# multiple architectures
+	charm_path = os.getenv('CHARM_HOME')
+
+	if (charm_path is None) and (home is not None):
+		if os.path.isdir(home + '/Charm/charm'):
```
brittonsmith, line 9: This function and `grackle_path_search` could be simplified a bit by creating a list of all possible paths, looping over them, and breaking when the right one is found.

config/stampede_intel.py
```diff
@@ -37,9 +52,22 @@
 #
 #--------------------------------------------------
 
-charm_path   = home + '/Charm/682/intel/omni/charm'
-papi_inc = home + '/include'
-papi_lib = home + '/lib'
+if os.path.isdir(home + '/Charm/682/intel/omni/charm'):
```
brittonsmith, line 55: `charm_path_search` and `grackle_path_search` could take arguments for additional paths to search first. That would eliminate the extra if statements.

aemerick:
This looks good to me, modulo if anything is changing given Britton's comments. I'll hold off on approval until I hear if anything will get changed there.

bwoshea:
Four approves!  Time to merge.

Issue #32, opened 2020-03-04
dir needs to be included in parameters in ICs for block_list


fearmayo:
As it stands the 'dir' parameter needs to be added to the output directive for the *.block_list files to be dumped. In the case where the dir is not specified the *.block_list files are not created.

mabruzzo:
This was fixed way back in PR #33

PR #33, opened 2020-03-05
Block list bugfix

aemerick:
Small bugfix allowing .block_list, .parameter, .file_list, and .libconfig files to be output in data-type outputs when no sub-directory is provided. In this case, the prefix to these files is adopted from the corresponding output file name. 

Fixes Issue #32


bwoshea:
@brittonsmith I'm adding you as a reviewer.  I will also add myself.

src/Cello/io_OutputData.cpp
```diff
@@ -133,57 +133,65 @@ void OutputData::write_block ( const  Block * block ) throw()
...
-  
-  if (name_dir != "") {
+  std::string name_file;
+  std::string name_out_file  = expand_name_(&file_name_,&file_args_);
+
+  if (name_dir == "") {
+    // output block list and parameters to work directory
+    name_dir  = ".";
+    // strip extension, use this for name
+    name_file = name_out_file.substr(0, name_out_file.find("."));
```
brittonsmith, line 145: Searching for the last period with `rfind` might be a bit more robust.

aemerick:
This should be good to go!

bwoshea:
Four approvals, merging!

PR #38, opened 2020-05-04
Fixed a bug that had prevented the use of the grackle cooling timestep.

mabruzzo:
The main issue was that `EnzoMethodGrackle`'s `timestep` instance method, declared in `src/Enzo/enzo_EnzoMethodGrackle.hpp` and implemented in `src/Enzo/enzo_EnzoMethodGrackle.cpp`, was not const-qualified. In contrast, the virtual method declared in the `Method` super-class (the one Cello uses to compute the timestep) is const-qualified. As a result, the old implementation had been providing an unused overload of the `timestep` method. When Cello was calling the `timestep` method of an instance of `EnzoMethodGrackle`, the default implementation provided for `Method::timestep` was used, which simply returned `std::numeric_limits<double>::max()`.

This bugfix can be tested by adding setting the Method::grackle::use_cooling_timestep parameter to true in `input/method_grackle.in`. Prior to this commit, changing this value will have no effect on the number of cycles that the simulation will run for. When the changes from this commit are included and the parameter is set to true, then the simulation will run for many,many more cycles.

### DIFF
--- a/src/Enzo/enzo_EnzoMethodGrackle.cpp
+++ b/src/Enzo/enzo_EnzoMethodGrackle.cpp
@@ -479,7 +479,7 @@ void EnzoMethodGrackle::compute_ ( EnzoBlock * enzo_block) throw()
 
 //----------------------------------------------------------------------
 
-double EnzoMethodGrackle::timestep ( Block * block ) throw()
+double EnzoMethodGrackle::timestep ( Block * block ) const throw()
 {
   const EnzoConfig * config = enzo::config();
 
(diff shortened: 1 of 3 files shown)


src/Enzo/enzo_EnzoMethodGrackle.cpp
```diff
@@ -512,17 +512,39 @@ double EnzoMethodGrackle::timestep ( Block * block ) throw()
...
+
+    // make sure to exclude the ghost zone. Including them can cause issues
+    // There is no refresh before this method is called (at least during the
+    // very first cycle) - this can lead to timesteps of 0 or smaller.
+
+    for (int iz = zstart; iz < zstop; iz++) {
+      for (int iy = ystart; iy < ystop; iy++) {
+        for (int ix = gx; ix < ngx - gx; ix++) {
+          dt = std::min(enzo_float(dt),
+                        std::abs(cooling_time[ix + ngx * (iy + ngy * iz)]) );
```
brittonsmith, line 544: Is there a macro that can be used to calculate the 1D array index? There's one in enzo, but I can't remember if there's one here.

aemerick: Yes there is (its used elsewhere in EnzoMethodGrackle): `INDEX(ix,iy,iz,ngx,ngy)`

aemerick: whoops. loooks like this got changed already

aemerick:
Looks good to me! Thanks for the fix

src/Enzo/enzo_EnzoMethodGrackle.cpp
```diff
@@ -512,17 +512,30 @@ double EnzoMethodGrackle::timestep ( Block * block ) throw()
...
     }
 
-    for (int i = 0; i < size; i++) dt = std::min(enzo_float(dt), std::abs(cooling_time[i]));
+    // make sure to exclude the ghost zone. Because there is no refresh before
+    // this method is called (at least during the very first cycle) - this can
+    // including ghost zones can lead to timesteps of 0
+    for (int iz = gz; iz < ngz - gz; iz++) {   // if rank < 3: gz = 0, ngz = 1
+      for (int iy = gy; iy < ngy - gy; iy++) { // if rank < 2: gy = 0, ngy = 1
+        for (int ix = gx; ix < ngx - gx; ix++) {
+	  int i = INDEX(ix, iy, iz, ngx, ngy);
```
brittonsmith, line 534: I'm not sure, but would it be better for the declaration of `i` to be outside the loop?

fearmayo: I suspect the compiler is going to take care of any optimisation issue here and keeping the variable scope within the nest is good practice.

mabruzzo: I very much share John's sentiments. I feel like it conveys more information about the use of `i`. I doubt it has any performance implications (especially since this is an int and not something that requires heap allocation - like a string).  I think this probably just boils down to a question of coding style.

brittonsmith: Ok, fair enough. It's fine with me.

brittonsmith: I've done a little reading on this and it does seem that this is a better way to do things. There are apparently additional optimizations the compiler can do when things are done this. Thanks!

bwoshea:
3 approves - merging!

Issue #37, opened 2020-05-01
EnzoMethodGrackle SMP-mode Error


mabruzzo:
I encountered an error while trying to run one of my branches in SMP-mode and the code crashed when `EnzoMethodGrackle::initialize_grackle_chemistry_data`  `initialize_chemistry_data`. Note that I have not attempted to replicate this on the master branch, but I believe that this is a general problem. I have attached the error message [here](https://pastebin.com/QXJXTxWY).

I intend to address this problem, but I could use a little input. 

@jobordner, could you confirm that there is a synchronization step between when the parameter files get read and when the method objects get initialized? If that is the case, then I am pretty sure that multiple PEs call `initialize_chemistry_data` at once and each try to initialize the same object in the global shared namespace.

I can think of 2 main approaches for solving this problem:
1. Introduce a global variable (or maybe a static member variable of each `EnzoMethodGrackle`) that tracks the most recent time at which the global chemistry data instance has been initialized and use some kind of locking mechanism to make sure only 1 PE per local node calls `initialize_chemistry_data(&grackle_units_)` per simulation timestep. The main downside to this approach is that I'm not sure what kind of locking mechanism to use. I've looked through the charm++ documentation a little and haven't come up with an obvious mechanism. @jobordner, are you aware of any approaches? (I don't suppose something like `std::lock` would work)
2. We could move away from using global data structures and track local versions of Grackle's chemistry data structures within each instance of `EnzoMethodGrackle.` To avoid needing to write a pup method for data structure, we could just avoid pupping the data structures (other than `chemistry_data`) altogether and simply reconstruct the data structures from scratch while unpacking.

@aemerick do you see any issues with either of these approaches or have any suggestions?

mabruzzo:
After looking at charm++'s documentation some more, I think a way to implement my first solution would be to introduce a NodeGroup object that provides exclusive entry methods for modifying the quantities in the global namespace of each logical node. 

In the absence of any alternative suggestions, I'm fairly sure I'm going to pursue my second solution.

mabruzzo:
This was resolved by PR #39

PR #39, opened 2020-05-11
Threadsafe Grackle Changes

mabruzzo:
This PR refactors calls to the external Grackle library to resolve SMP-mode issues.

**Problem Overview:**
The current procedure for setting up the `chemistry_data` struct on each PE (Processor Element) can be summarized by the following 2 steps:

  1. `EnzoConfig` dynamically allocates a local `chemistry_data` struct on each PE (Processor Element). It then calls `set_default_chemistry_parameters`, which both assigns this struct the default parameters and sets the global variable `grackle_data` equal to the address of this struct. Finally, the parameter file values are assigned to the appropriate fields of the struct that is pointed to by `grackle_data`. Note that the `EnzoConfig` destructor will deallocate the locally allocated struct.
  2. The constructor for `EnzoMethodGrackle` calls the instance method `EnzoMethodGrackle::initialize_grackle_chemistry_data`. This method ultimately executes  `initialize_chemistry_data`, which makes some modifications to the struct pointed to by the `grackle_data` global variable and sets up the `grackle_rates` global variable.
 
When running Enzo-E in non-SMP mode, each PE is on a separate 'logical' node. Thus has its own own copy of the global `grackle_data` and `grackle_rates` variables. However, in SMP-mode there are multiple PEs per 'logical' node and the global `grackle_data` and `grackle_rates` variables are shared between all PEs on a logical node.

This can leads to a few issues in SMP-mode:
  1. A small amount of memory can leak in `EnzoConfig` when multiple PEs try to store the result of `strdup` in `grackle_data->grackle_data_file`.
  2. There are memory leaks caused by the simultaneous calls to `initialize_chemistry_data` (pointers stored in `grackle_rates` that were allocated by one PE may be overwritten by the pointers allocated by another PE)
  3. It seems to be possible for the program to completely crash if the HDF5 files are simultaneously read by different PEs during `initialize_chemistry_data`. This is recorded in Issue #37 (I haven't been able to replicate it since I first reported it - so it seems like an infrequent occurrence).

**Solution:**
To resolve these issues, I have refactored the code to make use of the explicitly thread-safe local functions from grackle. In doing so, I generally tried to minimize changes to the existing functionallity. The most significant changes are as follows:
  - `EnzoMethodGrackle` now has a new private member variable, `grackle_rates` which stores the local instance of `chemistry_data_storage`. This is initialized in `EnzoMethodGrackle::initialize_grackle_chemistry_data`.
  - `EnzoComputePressure`, `EnzoComputeTemperature`, and `EnzoComputeCoolingTime` were refactored to call the new public instance methods of `EnzoMethodGrackle`: `calculate_pressure`, `calculate_temperature` and `calculate_cooling_time`. Doing this was cleaner than having each compute object retrieve the PE's local instances of `chemistry_data` and `chemistry_data_storage` from `EnzoConfig` and `EnzoMethodGrackle`. This change also makes `EnzoMethodGrackle` better encapuslate the majority of grackle-related operations. To help facillitate these changes, I introduced the `enzo::grackle_method` function to return a pointer to the PE's existing instance of `EnzoMethodGrackle` and new method to the `Problem` class to lookup registered `Method` instances by name (mirroring existing functionallity for `Physics` objects). 

Additionally, I moved the call to `EnzoMethodGrackle::initialize_grackle_chemistry_data` from the `compute` method to the `pup` method (during unpacking). Functionally, this only changes when the grackle structures are initialized after the object has been unpacked. This change removes all chances for the Compute objects to be called immediately after a (hypothetical) restart but before just means that the grackle_data can be initialized.

I have confirmed that this commit does not change the results of executing `input/method_grackle.in` (with just 1 PE or multiple PEs).

**Note:** I also considered (and partially implemented) an approach that used locks (`CmiNodeLock`)  and/or barrier synchronization (`CmiNodeBarrier`) to manage access to the global data structures. I ultimately choose to use the thread-safe local functions because it leads to simpler, less fragile code, has the same behavior in SMP and non-SMP mode, and is generally more consistent with the way other Methods are implemented. If there is a strong preference for using the global functions, I can revisit the approach.

In the future, (if these changes are approved) it may be worth creating a physics object to encapsulate all grackle operations (or maybe more generally to encapsulate equation-of-state related operations - I had already started thinking about this sort of thing with regards to my VL+CT integrator PR)

### DIFF
--- a/src/Enzo/enzo_EnzoConfig.cpp
+++ b/src/Enzo/enzo_EnzoConfig.cpp
@@ -168,8 +168,10 @@ EnzoConfig::EnzoConfig() throw ()
 EnzoConfig::~EnzoConfig() throw ()
 {
 #ifdef CONFIG_USE_GRACKLE
-  delete[] method_grackle_chemistry->grackle_data_file;
-  delete method_grackle_chemistry;
+  if (method_grackle_chemistry){
+    delete[] method_grackle_chemistry->grackle_data_file;
+    delete method_grackle_chemistry;
+  }
 #endif // CONFIG_USE_GRACKLE
 }
 
--- a/src/Enzo/enzo_EnzoConfig.cpp
+++ b/src/Enzo/enzo_EnzoConfig.cpp
@@ -168,6 +168,7 @@ EnzoConfig::EnzoConfig() throw ()
 EnzoConfig::~EnzoConfig() throw ()
 {
 #ifdef CONFIG_USE_GRACKLE
+  free(method_grackle_chemistry->grackle_data_file);
   delete method_grackle_chemistry;
 #endif // CONFIG_USE_GRACKLE
 }
@@ -340,10 +341,7 @@ void EnzoConfig::pup (PUP::er &p)
       method_grackle_chemistry->use_grackle = method_grackle_use_grackle;
 
       if(method_grackle_chemistry->use_grackle){
-        if (set_default_chemistry_parameters(method_grackle_chemistry) == ENZO_FAIL){
-          ERROR("EnzoMethodConfig::pup()",
-                "Error in Grackle set_default_chemistry_parameters");
-        }
+        *method_grackle_chemistry = _set_default_chemistry_parameters();
       }
     }
   }
@@ -808,17 +806,13 @@ void EnzoConfig::read(Parameters * p) throw()
   if (method_grackle_use_grackle) {
 
     method_grackle_chemistry = new chemistry_data;
-
-    if (set_default_chemistry_parameters(method_grackle_chemistry) == ENZO_FAIL) {
-      ERROR("EnzoMethodGrackle::EnzoMethodGrackle()",
-      "Error in set_default_chemistry_parameters");
-    }
+    *method_grackle_chemistry = _set_default_chemistry_parameters();
 
     /* this must be set AFTER default values are set */
     method_grackle_chemistry->use_grackle = method_grackle_use_grackle;
 
     // Copy over parameters from Enzo-P to Grackle
-    grackle_data->Gamma = field_gamma;
+    method_grackle_chemistry->Gamma = field_gamma;
 
     //
     method_grackle_use_cooling_timestep = p->value_logical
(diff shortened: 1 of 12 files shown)


bwoshea:
@mabruzzo could you please resolve the merge conflicts in this PR?  I've merged PR #38 , which should hopefully make this easier.

bwoshea:
@brittonsmith @jwise77 , would you mind taking a look at this PR after @mabruzzo fixes the merge conflict?

mabruzzo:
I solved the merge conflicts and re-confirmed that the results from running input/method_grackle don't change. This is now ready for review pending the completion of the CI test-suites

src/Enzo/enzo_EnzoConfig.cpp
```diff
@@ -168,6 +168,7 @@ EnzoConfig::EnzoConfig() throw ()
 EnzoConfig::~EnzoConfig() throw ()
 {
 #ifdef CONFIG_USE_GRACKLE
+  free(method_grackle_chemistry->grackle_data_file);
```
jwise77, line 171: Why free() and not delete? This string was allocated with new.

src/Enzo/enzo_EnzoConfig.cpp
```diff
@@ -340,10 +341,7 @@ void EnzoConfig::pup (PUP::er &p)
       method_grackle_chemistry->use_grackle = method_grackle_use_grackle;
 
       if(method_grackle_chemistry->use_grackle){
-        if (set_default_chemistry_parameters(method_grackle_chemistry) == ENZO_FAIL){
-          ERROR("EnzoMethodConfig::pup()",
-                "Error in Grackle set_default_chemistry_parameters");
-        }
+        *method_grackle_chemistry = _set_default_chemistry_parameters();
```
aemerick, line 344: Looking at this code again, I think this should probably be changed to 

```
method_grackle_chemistry = new chemistry_data;
if (method_grackle_use_grackle){
  *method_grackle_chemistry = _set_default_chemsitry_parameters();
}
method_grackle_chemistry->use_grackle = method_grackle_use_grackle;
```
Since `_set_default_chemistry_parameters()` resets `use_grackle` to `FALSE`. This issue is there in the previous version too, so I'm not too sure why this didn't cause problems.

aemerick: This is done correctly in `EnzoConfig::read`

src/Enzo/enzo_EnzoConfig.cpp
mabruzzo: It's because the grackle_data_file is allocated by `strdup` which I believe uses `malloc` internally. I'll double-check this and add a comment documenting this.

src/Enzo/enzo_EnzoConfig.cpp
mabruzzo: Thanks for catching this! I don't think this code has ever actually been tested because I believe that `EnzoConfig::pup` is only used for restarts. However, I think this wasn't actually a problem since the subsequent `p | *method_grackle_chemistry;` overwrites all of the attributes anyway. 

Now that I'm thinking about it, I the `set_default_chemistry_parameters(method_grackle_chemistry);` in the earlier version of the `pup` method was only necessary to set the value of the `grackle_data` global variable.  This means it isn't necessary to call `_set_default_chemsitry_parameters();`.

While we are on the topic, do you remember if there is a reason why we explicitly compute the `is_null` variable? It seems redundant with the value of `!method_grackle_chemistry`.

aemerick: Ah ok. That makes sense. So we can safely get rid of the call to `_set_default_chemistry_parameters()` here then? 

I vaguely remember the `is_null` variable being either from @jobordner or something I had copied based on similar code elsewhere. But I can't find the latter now and looking at the log, this bit was put in by @jobordner . Looking at the Charm++ documentation this is the way this should be done if the pointer is potentially NULL (https://charm.readthedocs.io/en/latest/charm++/manual.html#null-object-pointer)

mabruzzo:
The most recent commit should addresses both Andrew's and John's comments. 

This is ready for further review

jobordner:
A lot of regression tests crash on exit from trying to dereference a null pointer.  This is easily fixed with the following changes to `enzo_EnzoConfig.cpp`:

```
 EnzoConfig::~EnzoConfig() throw ()
 {
 #ifdef CONFIG_USE_GRACKLE
-  delete[] method_grackle_chemistry->grackle_data_file;
-  delete method_grackle_chemistry;
+  if (method_grackle_chemistry) {
+    delete[] method_grackle_chemistry->grackle_data_file;
+    delete method_grackle_chemistry;
+  }
 #endif // CONFIG_USE_GRACKLE
 }

```

bwoshea:
4 approvals, merging!

PR #42, opened 2020-06-04
Color fields to refresh in PPm

aemerick:
Adding colour (which will need to be changed to color once my other PR gets accepted -- i think --) fields to refresh list in PPM. 

I figured this would be useful for the hack week.


aemerick:
I think it may be worth waiting on this until solver-dd gets pulled in since that changes the refresh a bit (and will actually simplify this quite a bit).

bwoshea:
@aemerick the solver-dd PR is now pulled in.  Does anything else have to happen to this PR?

bwoshea:
Note: as per my discussion with @aemerick , I believe we are going to scrap this PR because its content is included in PR #49 .  However, since that is still pending we are going to defer merging or declining this PR for the time being.  @aemerick can you confirm?

jwise77:
Another PR to close since it's included in #49

PR #46, opened 2020-06-09
use installed scons for build

brittonsmith:
This is an attempt at using Python 3 and an installed version of scons for the build. Currently, we use a local copy of an older version of scons which does not support Python 3. I did some pretty quick and dirty python2->python3 conversion that I can clean up if this works and we like it.


brittonsmith:
@mabruzzo has verified that `scons` is easy to install via `pip` on Frontera, so requiring an external installation should like not be a problem elsewhere.

If I don't hear any objections to this PR in principle, then I'll go ahead and do a little more cleanup and make it ready for review.

brittonsmith:
@bwoshea, it's minimally ready. I was hoping to find some time to make some of the python2->3 conversion stuff look a bit cleaner, but I don't think I'm going to for a while. None of that is essential, so I'd say to go ahead and review/merge when ready.

forrestglines:
@brittonsmith This builds for me (after installing scons) when grackle is enabled, but fails when grackle is disable in `SConstruct`. The `master` branch builds fine without grackle, so it's probably something that needs to be merged in from master into this branch.

Also, we should add installation of `scons` to the build tutorial in `doc/source/tutorial/getting_started.rst`.

forrestglines (approved):
I think this looks good. The one catch I see would be to add some documentation on the "Getting Started" tutorial for Comet. Installing SCons via pip3 in the user space would be the standard approach there, right?

bwoshea:
@brittonsmith some of the tests are failing for this PR.  Would you mind looking at that?  Otherwise, I think we need one more approval and we're good to go.

mabruzzo (approved):
This looks great to me!

config/_common_search_paths.py
```diff
@@ -1,37 +1,36 @@
 import os
 
 def _check_dirs(search_paths):
-        for search_path in search_paths:
-                if os.path.isdir(search_path):
-                        return search_path
-        return None
+    for search_path in search_paths:
+        if os.path.isdir(search_path):
+            return path
```
gregbryan, line 6:
```suggestion
            return search_path
```
I think this undoes a change @mabruzzo made in PR#65.

gregbryan (commented):
One suggested correction, otherwise looks good to me.

config/_common_search_paths.py
brittonsmith: Oh yes, thanks. That was unintentional.

gregbryan (approved):

bwoshea:
3 approves, this PR is getting merged!

PR #47, opened 2020-06-09
Add name to job commands.

brittonsmith:
This doesn't change any commands, just adds a name to each section of the build. The only thing this does is make it so the steps on circleci have easily parsable names instead of a word salad of all the commands being run.


jwise77:
Three approvals. I'm merging.

PR #45, opened 2020-06-09
Fixes for restart bugs (issue #5)

jobordner:
This PR addresses two bugs associated with Issue #5:

1. corrupted strings written to parameters.libconfig files after a restart (bug fix by Matthew Abruzzo) 
2. EnzoBlock attribute values not getting written to HDF5 files after restart.

PR #51, opened 2020-06-30
Pup-restart-bugfixes

mabruzzo:
Fixed some bugs in pup routines that manifest during restarts. These bugs include:

- improper pupping of a pointer in `BoundaryValue::pup`
- fields of `chemistry_data` were not being pupped in `void operator|(PUP::er &p, chemistry_data &c)`. These are probably new additions present in more recent versions of Grackle
- issues in recomputing the internal grackle data structures in `EnzoMethodGrackle::pup` after reconstruction (I accidentally introduced these issues in PR #39). This can be broken into two parts:
    1. In the pup routine, the value assigned to the `time_grackle_data_initialized_` attribute immediately preceding the call to `initialize_grackle_chemistry_data` caused the function to exit early before initialize any of grackle's internal data structures
    2. During the restart, the fact that `Method` objects are unpacked before `Unit` objects caused issues when `initialize_grackle_chemistry_data` attempted to initialize the `grackle_units_` attribute.  This was easy to resolve since it wasn't actually necessary to initialize `grackle_units_` when unpacking (the state was already being packed up/unpacked within `EnzoMethodGrackle::pup`)

**EDIT:** I figured this PR would be most place to appropriate to introduce my `ckpt_restart_test.py` testing tool, since this PR it has not yet been reviewed and I've used the testing tool to introduce tests that prevent these changes from regressing in the future. Note that these tests will all fail before PR #45 gets merged in - however I have confirmed locally that these tests will all pass once those changes are included.

`ckpt_restart_test.py` is a command line tool designed to assist with running checkpoint-restart tests. I choose to implement this because many of the PUP routines have minimal coverage, and the existing PPM and PPML tests require a lot of boilerplate code to setup (this tool makes it much easier to setup tests for new methods). I waffled a little between having SCons directly invoke the tool (as a python function) or call it as a command line tool. I choose the latter so that the test is easy to re-run without invoking the entire test suite.

The tool takes a Cello/Enzo-E input file, a value called `RESTART_CYCLE`, and a value called `STOP_CYCLE`. Then it performs the following procedure:
  1. The specified Cello/Enzo-E input file is copied and modified such that the simulation writes a checkpoint output to disk at `RESTART_CYCLE`, runs until `STOP_CYCLE` and writes a data output including all fields at `STOP_CYCLE`.
  2. The simulation is run using the modified input file and all outputs are moved to a separate directory (a copy of the checkpoint is left in the original directory).
  3. The simulation is then restarted and the new outputs are moved to a separate directory.
  4. Finally, the program ensures that the output files written at `STOP_CYCLE` contain identical data before and after the restart. This is done in a way that accounts for the fact that block data can be stored to hdf5 files in an arbitrary order.

The program writes the result to an output file that can be parsed by the testing framework (this is implemented by `tools/test_report.py`). This program is slightly more complicated than it needs to be, but it gets the job done. Once PR #46 is pulled in, I plan to make the VL+CT tests reuse some of the infrastructure introduced here. Note that this tool does require that the h5diff utility is installed. I tweaked `.circleci/config.yml` to make sure that this tool is installed.

I introduced the following 3 input files and tweaked `tools/Restart/SConscript` to automatically use them in tests.
- `input/Checkpoint/checkpoint_vlct.in`
- `input/Checkpoint/checkpoint_boundary.in` (Directly tests the bug fixed in this PR)
- `input/Checkpoint/checkpoint_grackle.in` (Directly tests the bug fixed in this PR)

Some extra changes were necessary to use the grackle method in the tests (because the path to the Grackle data files varies between systems). In short, the `GRACKLE_INPUT_DATA_DIR` environment variable is used to specify the path to the directory holding the Grackle input data files. The `test/Restart/SConscript` file instructs SCons to use that information to writes a short input file that includes the contents of `input/Checkpoint/checkpoint_grackle.in` and overwrites the path to the data file.


mabruzzo:
I rebased this branch after PR #45 got merged (since nobody is using this). The tests should now all pass

bwoshea:
@jobordner would you please take a look at this PR?

jobordner (approved):

mabruzzo:
@bwoshea This was the PR I was talking about. If you could assign a second person to review this that would be great! 

This is relatively important for restarting any simulations involving Grackle

clairekope (approved):

clairekope (approved):
Look good to me, and James already approved

bwoshea:
@mabruzzo would you mind fixing the merge conflicts listed here?  If so, we can go ahead and merge.  Thank you!

gregbryan (approved):
Looks good to me once conflict is fixed.

bwoshea:
3 approvals and all tests pass.  I'm going to merge this!

PR #53, opened 2020-08-07
Testing framework

SaoirseWard:
The test and input directories have been restructured in order for it to be easier to create quick-to-run tests. The test subdirectories and what they contain was based on the original test/SConscript file and how it broke up the testing components. build.sh has been edited to count .unit in the test child directories as opposed to the test directory. The test/SConscript file has been edited to call each SConscript file in the child directories. Each test subdirectory has an SConscript file to call the tests in the directory. Each input file has been checked to determine if it is working and the tests that do not work have been moved to the Obsolete subdirectory. The other .in  files have been sorted based on what they test. Documentation has been written to explain each test as well as the testing infrastructure and how to create a new test.


fearmayo:
@SaoirseWard there are a few typos in the "The Testing Environment" page. Could you have a look and squash these?

PR #63, opened 2020-11-19
TESTPR: Circleci vlct integrator

mabruzzo:
This is to test whether we can revert the default optimization flag from `-O1` back to `-O3` if we update the circle configuration. This is not actually meant to be reviewed separately from #62 . The point is to avoid polluting the commit history while testing different circleci settings.


.circleci/config.yml
```diff
@@ -39,6 +39,15 @@ commands:
               cd charm-v$CHARM_VER
               ./build charm++ netlrts-linux-x86_64 -j4 --with-production
             fi
+            # prepare testing dependencies
+            python3 -m venv $HOME/venv
+            source $HOME/venv/bin/activate
+            pip install --upgrade pip
+            pip install --upgrade wheel
+            pip install --upgrade setuptools
+            # matplotlib 3.3.0 breaks yt 3.6.0 (latest stable version at this
```
brittonsmith, line 48: yt 3.6.1 was released just a couple days ago, so we should no longer need this. That said, experience has shown it's probably wise to continue to pin the matplotlib version.

mabruzzo:
These changes have successfully been added into the main PR (#62)

PR #65, opened 2021-01-07
Added a frontera compilation target.

mabruzzo:
I'm not sure if we want to unify the Frontera and Stampede targets, but I wanted to get this process started (I've had these changes sitting around for a long time).

This PR also fixes a minor bug in `config/_common_search_paths.py`


mabruzzo:
To be clear, I think that this is fine to pull into the main branch. But, if any of the reviewers thing we should consolidate the frontera and stamepede targets let me know and I can iterate

jwise77 (approved):

bwoshea:
@mabruzzo  I don't think we need to integrate the config files for the two computers.  This seems perfectly reasonable to me!

bwoshea (approved):

gregbryan (approved):

PR #73, opened 2021-01-08
slimmed down testing file for debugging

fearmayo:
Debug why tests are failing on circle-ci


bwoshea:
@fearmayo could you resolve the conflict?  Also, would you mind commenting on why the tests were failing, so we have documentation of it?

fearmayo:
> @fearmayo could you resolve the conflict? Also, would you mind commenting on why the tests were failing, so we have documentation of it?

Just resolving the conflict now. However, I could not fix the issue. The tests are failing due to an issue with ssh on circle-ci. They run ok on my local cluster. It's a setup issue. Maybe @mabruzzo or @brittonsmith know what's going wrong?

mabruzzo:
So I've logged into circleci and played around with things, and I think I've figured out the problem.

First a little context. As you probably know, we can't ssh on circleci (although maybe it would be possible to setup passwordless ssh for localhost?), so we need to the command line form of parallel runs from
`charmrun +p4 bin/enzo-p <input_file>`
to 
`charmrun +p4 bin/enzop ++local <input_file>`

In short, the reason the reason that the test fails on circleci is due to a minor oversight during the testing overhaul that doesn't make use of the machinery to modify the command line arguments in this way. That's also why the tests successfully runs on your local cluster work, (because it has working ssh - or you compiled charm++ with mpi).

In more detail, the `++local` is included in the command line arguments as follows:
- we set the environment variable `CHARM_ARGS` to have a value of `++local`. (This is done on line 13 of `.circleci/config.yml`)
- On lines 19 and 20 of `test/SConscript` this value is read in from the environment variable and stored in the `parallel_args` variable.
- Then the `parallel_args` variable is used while constructing the `RunParallel` `Builder` object on line 28 of `test/SConscript`.

Prior to the testing overhaul, this `RunParallel` builder was used to execute all parallel test runs. However, during the overhaul, this was changed. 

For example, within `test/AmrPpm/SConscript`, the `"RunAdapt"` builder is now used. This builder is constructed (on line 29 of `test/AmrPpm/SConscript`). It's construction is similar to `RunParallel`, but it doesn't include the value of the `parallel_arg` variable in the command line argument.

Once you resolve this, all of the tests should pass on circleci.

fearmayo:
> So I've logged into circleci and played around with things, and I think I've figured out the problem.
> 
> First a little context. As you probably know, we can't ssh on circleci during the tests, so we need to the command line form of parallel runs from
> `charmrun +p4 bin/enzo-p <input_file>`
> to
> `charmrun +p4 bin/enzop ++local <input_file>`
> 
> In short, the reason the reason that the test fails on circleci is due to a minor oversight during the testing overhaul that doesn't make use of the machinery to modify the command line arguments in this way. That's also why the tests successfully runs on your local cluster work, (because it has working ssh - or you compiled charm++ with mpi).
> 
> In more detail, the `++local` is included in the command line arguments as follows:
> 
>     * we set the environment variable `CHARM_ARGS` to have a value of `++local`. (This is done on line 13 of `.circleci/config.yml`)
> 
>     * On lines 19 and 20 of `test/SConscript` this value is read in from the environment variable and stored in the `parallel_args` variable.
> 
>     * Then the `parallel_args` variable is used while constructing the `RunParallel` `Builder` object on line 28 of `test/SConscript`.
> 
> 
> Prior to the testing overhaul, this `RunParallel` builder was used to execute all parallel test runs. However, during the overhaul, this was changed.
> 
> For example, within `test/AmrPpm/SConscript`, the `"RunAdapt"` builder is now used. This builder is constructed (on line 29 of `test/AmrPpm/SConscript`). It's construction is similar to `RunParallel`, but it doesn't include the value of the `parallel_arg` variable in the command line argument.
> 
> Once you resolve this, all of the tests should pass on circleci.

Hi @mabruzzo ,

Great - thanks! I had wondered about inserting the ++local argument. So from my reading of your comment it looks like `RunAdapt` also needs to include the `parallel_arg` arguments?

mabruzzo:
> > So I've logged into circleci and played around with things, and I think I've figured out the problem.
> > First a little context. As you probably know, we can't ssh on circleci during the tests, so we need to the command line form of parallel runs from
> > `charmrun +p4 bin/enzo-p <input_file>`
> > to
> > `charmrun +p4 bin/enzop ++local <input_file>`
> > In short, the reason the reason that the test fails on circleci is due to a minor oversight during the testing overhaul that doesn't make use of the machinery to modify the command line arguments in this way. That's also why the tests successfully runs on your local cluster work, (because it has working ssh - or you compiled charm++ with mpi).
> > In more detail, the `++local` is included in the command line arguments as follows:
> > ```
> > * we set the environment variable `CHARM_ARGS` to have a value of `++local`. (This is done on line 13 of `.circleci/config.yml`)
> > 
> > * On lines 19 and 20 of `test/SConscript` this value is read in from the environment variable and stored in the `parallel_args` variable.
> > 
> > * Then the `parallel_args` variable is used while constructing the `RunParallel` `Builder` object on line 28 of `test/SConscript`.
> > ```
> > 
> > 
> > Prior to the testing overhaul, this `RunParallel` builder was used to execute all parallel test runs. However, during the overhaul, this was changed.
> > For example, within `test/AmrPpm/SConscript`, the `"RunAdapt"` builder is now used. This builder is constructed (on line 29 of `test/AmrPpm/SConscript`). It's construction is similar to `RunParallel`, but it doesn't include the value of the `parallel_arg` variable in the command line argument.
> > Once you resolve this, all of the tests should pass on circleci.
> 
> Hi @mabruzzo ,
> 
> Great - thanks! I had wondered about inserting the ++local argument. So from my reading of your comment it looks like `RunAdapt` also needs to include the `parallel_arg` arguments?

That's exactly right.

From skimming through the test files, this looks like this type of change is necessary for other test runners too (e.g. `RunPpm8`, `RunCosmology_8`, etc.).

I don't know a lot about SCons, but it could potentially be easier to try to reuse the `RunParallel` builder wherever possible (you could probably do that in most places other than the load balancing tests)

fearmayo:
> > > So I've logged into circleci and played around with things, and I think I've figured out the problem.
> > > First a little context. As you probably know, we can't ssh on circleci during the tests, so we need to the command line form of parallel runs from
> > > `charmrun +p4 bin/enzo-p <input_file>`
> > > to
> > > `charmrun +p4 bin/enzop ++local <input_file>`
> > > In short, the reason the reason that the test fails on circleci is due to a minor oversight during the testing overhaul that doesn't make use of the machinery to modify the command line arguments in this way. That's also why the tests successfully runs on your local cluster work, (because it has working ssh - or you compiled charm++ with mpi).
> > > In more detail, the `++local` is included in the command line arguments as follows:
> > > ```
> > > * we set the environment variable `CHARM_ARGS` to have a value of `++local`. (This is done on line 13 of `.circleci/config.yml`)
> > > 
> > > * On lines 19 and 20 of `test/SConscript` this value is read in from the environment variable and stored in the `parallel_args` variable.
> > > 
> > > * Then the `parallel_args` variable is used while constructing the `RunParallel` `Builder` object on line 28 of `test/SConscript`.
> > > ```
> > > 
> > > 
> > > Prior to the testing overhaul, this `RunParallel` builder was used to execute all parallel test runs. However, during the overhaul, this was changed.
> > > For example, within `test/AmrPpm/SConscript`, the `"RunAdapt"` builder is now used. This builder is constructed (on line 29 of `test/AmrPpm/SConscript`). It's construction is similar to `RunParallel`, but it doesn't include the value of the `parallel_arg` variable in the command line argument.
> > > Once you resolve this, all of the tests should pass on circleci.
> > 
> > 
> > Hi @mabruzzo ,
> > Great - thanks! I had wondered about inserting the ++local argument. So from my reading of your comment it looks like `RunAdapt` also needs to include the `parallel_arg` arguments?
> 
> That's exactly right.
> 
> From skimming through the test files, this looks like this type of change is necessary for other test runners too (e.g. `RunPpm8`, `RunCosmology_8`, etc.).
> 
> I don't know a lot about SCons, but it could potentially be easier to try to reuse the `RunParallel` builder wherever possible (you could probably do that in most places other than the load balancing tests)

Cool - thanks @mabruzzo

fearmayo:
The tests are running now. I had to manually add the "++local" variable to each of the `Builder` calls. It could probably be further optimised in the future but for now I think this approach is sufficient.

test/Balance/SConscript
```diff
@@ -1,41 +1,43 @@
...
 Import('test_path')
-
+import os
 #----------------------------------------------------------------------
 # defines
 #----------------------------------------------------------------------
 
 env['CPIN'] = 'touch parameters.out; mv parameters.out ${TARGET}.in'
 env['RMIN'] = 'rm -f parameters.out'
+env['clocal_cmd'] = '++local'
```
mabruzzo, line 15: More generally, it may make more sense to simply skip these tests on circleci, rather than adding the `++local` argument, since this explicitly is checking on load-balancing between processes. I think that this is how these tests were handled prior to the testing overhaul (but I may be mistaken).

If doing this would be difficult and time-consuming, it would be fine with me if you just leave things as they stand and open an Issue mentioning that this should be addressed.

test/AmrPpm/SConscript
```diff
@@ -26,9 +26,9 @@ env.Append(BUILDERS = { 'pngToGif'  : png_to_gif})
 
 
 
-run_adapt = Builder(action = "$RMIN; " + date_cmd + parallel_run + " $SOURCE $ARGS " + " > $TARGET 2>&1; $CPIN; $COPY")
+run_adapt = Builder(action = "$RMIN; " + date_cmd + parallel_run + " $SOURCE $clocal_cmd  $ARGS " + " > $TARGET 2>&1; $CPIN; $COPY")
 env.Append(BUILDERS = { 'RunAdapt' : run_adapt } )
-env_mv_adapt = env.Clone(COPY = 'mkdir -p ' + test_path + '/AmrPpm/Adapt-L5-P1; mv `ls *.png *.h5` ' + test_path + '/AmrPpm/Adapt-L5-P1')
```
mabruzzo, line 31: I just wanted to confirm that the removal of the second part of this command is intentional. 

Lot's of other environments seem to have that second part of the command (like `env_mv_heat_8` from `test/MethodHeat/SConscript`).

build.sh
```diff
@@ -236,7 +237,10 @@ if [ $target == "test" ]; then
     file_started=test/runs_started.$configure
     file_completed=test/runs_completed.$configure
 
+    ls test/*/test_*.unit 
     ls test/*/test_*.unit                   > $file_attempted
+    head test/*/test_*.unit
+    grep  "BEGIN" test/*/test_*.unit
```
mabruzzo, line 243: Were these 3 new lines just for debugging? Or are they meant to be permanent additions?

test/SConscript
```diff
@@ -47,10 +41,16 @@ env_rm_png  = env.Clone(COPY = 'rm -f *.png')
...
+
+#----------------------------------------------------------------------
+SConscript('Balance/SConscript')
+#----------------------------------------------------------------------
+
 #----------------------------------------------------------------------
 # ARRAY COMPONENT         
 #----------------------------------------------------------------------
-env.RunSerial('test_CelloArray.unit',bin_path + '/test_CelloArray')
+#env.RunSerial('test_CelloArray.unit',bin_path + '/test_CelloArray')
```
mabruzzo, line 52: This test got added during PR #62. It must have fallen through the cracks during the merging process.

What do we need to do to get this test running? Do I just need to setup an `ArrayComponent/SConscript` file and move this line to that file?

mabruzzo (changes requested):
I think this looks pretty good. I agree that the manual addition of `++local` is probably good enough for the short term (to make sure that the test suite gives a reasonable response).

We might want to open an Issue on this topic to make sure it doesn't get lost.

I just have a few minor questions/comments. Other than those, I think this is good to go.

test/SConscript
fearmayo: Yes - exactly. I think you can follow what is done in MethodPpm  - that appears to do something similar. Maybe once this PR gets pulled in you can issue that PR?

build.sh
fearmayo: fixed

test/AmrPpm/SConscript
fearmayo: I've put this back in now

test/Balance/SConscript
fearmayo: I think I will open an issue

test/Balance/SConscript
```diff
@@ -1,41 +1,43 @@
 Import('env')
 Import('parallel_run')
+Import('parallel_arg')
```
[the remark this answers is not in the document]
fearmayo: done

test/Balance/SConscript
```diff
@@ -1,41 +1,43 @@
...
 #env.Append(BUILDERS = { 'RunBalanceHybrid' : run_balance_hybrid } )
 #env_mv_hybrid  = env.Clone(COPY = 'mkdir -p ' + test_path + '/Balance/Hybrid; mv `ls *.png *.h5` ' + test_path + '/Balance/Hybrid')
 
-run_balance_rand_cent = Builder(action = "$RMIN; " + date_cmd + parallel_run + " $SOURCE $ARGS +balancer RandCentLB > $TARGET 2>&1; $CPIN; $COPY")
+run_balance_rand_cent = Builder(action = "$RMIN; " + date_cmd + parallel_run + " $SOURCE  $clocal_cmd $ARGS +balancer RandCentLB > $TARGET 2>&1; $CPIN; $COPY")
 env.Append(BUILDERS = { 'RunBalanceRandCent' : run_balance_rand_cent } )
 env_mv_rand_cent  = env.Clone(COPY = 'mkdir -p ' + test_path + '/Balance/RandCent; mv `ls *.png *.h5` ' + test_path + '/Balance/RandCent')
 
-run_balance_refine = Builder(action = "$RMIN; " + date_cmd + parallel_run + " $SOURCE $ARGS +balancer RefineLB > $TARGET 2>&1; $CPIN; $COPY")
+run_balance_refine = Builder(action = "$RMIN; " + date_cmd + parallel_run + " $SOURCE   $clocal_cmd $clocal_cmd $ARGS +balancer RefineLB > $TARGET 2>&1; $CPIN; $COPY")
```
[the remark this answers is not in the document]
fearmayo: Thanks!

test/SConscript
mabruzzo: Sure, will do.

mabruzzo (approved):

gregbryan (approved):

bwoshea:
OK, we've got two approves, I'm merging!

Issue #69, opened 2021-01-07
Deprecated Code introduced by VL+CT PR


mabruzzo:
Some deprecated code was introduced by the VL+CT PR (#62)

- One of those spots was in `EnzoInitialCloud`
- Another spot was in `EnzoFieldArrayFactory`

These need to be removed

PR #74, opened 2021-01-11
Enable VLCT to run without Bfields and transition implmentation from using Groupings to Maps

mabruzzo:
## Summary

This Pull Request makes introduces 3 key changes to the VL+CT integrator.

1. Magnetic fields can now be switched off.
    - A new `mhd_choice` parameter has been added for the VL+CT integrator to indicate whether Bfields should be included. This parameter has no default value - an error will be raised if it isn't specified.
    - To actually make this change, we refactored the `EnzoConstrainedTransport` class. The class now completely encapsulates handling of the interface bfields.
        - I'm not sold on the current design of `EnzoConstrainedTransport`.
        - Currently the class encapsulate the interface bfield and has mutable state; we need to initialize a new instance each time we call `EnzoMethodMHDVlct::compute`. I think a better design would separate the actual constrained transport operations and the responsibility of tracking the extra data used in these operations to separate classes.
        - If we ever add support for other MHD update approaches (e.g. divergence cleaning or other constrained transport methods), we'll probably want to make `EnzoConstrainedTransport` a subclass of some `EnzoBfieldUpdate` class to allow for easy swapping. Along those lines, it might also make sense to introduce a dummy Bfield updator class that is used when we don't want any bfields.
        - For now, I'm inclined to stick current design of `EnzoConstrainedTransport` and wait until a better design becomes apparent or necessary. However, if anybody has better ideas, do let me know.
    - I modified some tests that were previously used with magnetic fields set
      to 0 to now test the integrator without any magnetic fields at all.

2. Transitioned `EnzoMethodMHDVlct`, `EnzoEquationOfState`, `EnzoRiemann`, `EnzoReconstructor`, `EnzoIntegrableUpdate`, and `EnzoConstrainedTransport` away from using `Grouping`s of fields to directly employing maps/dictionaries that hold `EFlt3DArray`s. This transition affords a number of benefits:
   - This transition affords a number of benefits:
       - It simplifies the code. Our usage of `Grouping`s was already conceptually similar to a mapping, but it has an extra level of indirection.
       - It decouples the hydro algorithms from the Cello machinery.
       - Operations can be expressed more concisely. Our prior implementation exercised `Grouping`s in a manner that probably wasn't initially anticipated and likely resulted in more verbose code.
       - See the **Future Work** heading for a list of future improvements that I plan to pursue by leveraging this newer functionality
   - To help facilitate this transition we introduced the `EnzoEFlt3DArray` class which serves as a map of `EFlt3DArray`s. The implementation is fairly naive and unoptimized. I discuss potential improvements below under the **Future Work** heading.
   - This transition allowed me to drastically reduce the number of methods of `EnzoFieldArrayFactory.` In the process I was able to removed the deprecated method mentioned by Issue #69. At this point only a single non-constructor method `EnzoFieldArrayFactory::from_name`. This method builds an array that wraps a given Cello field. A future PR should probably move this functionallity to Cello's `Field` class and remove `EnzoFieldArrayFactory` completely

3. Shifted from pre-registerring the passively advected scalars in `EnzoRiemann`, `EnzoReconstructor`, `EnzoIntegrableUpdate`, and `EnzoEquationOfState` to specifying them when executing their functionallity
   - It was not originally my intention to include this change in this PR, but it kind of sneaked into it.
   - This has always been the better approach to allow for advection of passive scalars that are only registered by `Method` objects after the `EnzoMethodMHDVlct` is initialized or by subclasses of `Initial`. The (pending) introduction of the required field machinery makes this all the more important.
   - To help facilitate this change, we introduced the `EnzoNestedPassiveScalarFieldList` class to generate and cache the list of passively advected scalars. While introducing a new class to handle this isn't strictly necessary, this is a general challenge faced by all (M)HD integrators.
   - To dynamically specify specify the passively advected scalars to the integration components, we now commonly pass around a `passive_lists` argument.
       - This is initialized by `EnzoNestedPassiveScalarFieldList`.
       - It's a vector of vectors of passive scalar names (they serve as keys for the maps of arrays). The first list holds the name of all passively advected scalars that are normally advected (e.g. enzo colours). All subsequent lists hold collections of passively advected scalars whose mass fraction must sum to 1 (e.g. enzo species). There may be a better format for conveying this information, but this seemed like a reasonable way to start.
       - *Note:* A future PR still needs to add support to `EnzoMethodMHDVlct` for advecting collections of passive scalars, like species, whose mass fractions must sum to 1.
   - I also found it necessary to change `EnzoMethodMHDVlct` so that it refreshes all fields (because all fields holding passively advected scalars are not known at compile time)

Apologies about how wide reaching these changes are. It was initially my intention to just perform changes 1 and 2 (because they both will require moderate changes to the documentation). If I had been more careful, the last change could have been introduced as a separate PR, but some of the changes definitely eased the transition.

## Future Work

I just wanted to briefly mention some future changes I anticipate making thanks to this transition from `Grouping`s to `Map`s:

- I'm going to attempt to refactor `EnzoComputePressure` (and the relevant machinery in `EnzoMethodGrackle`) to make use of the `EnzoEFltArrayMap` machinery. I think this should be relatively easy to do while maintaining an interface that falls back to using the Cello fields. If successful, this will facilitate adding full Grackle support to `EnzoEquationOfState` and `EnzoMethodMHDVlct`
- It's my intention to introduce some kind of memory pool so that we don't have to allocate and deallocated memory every time we execute `EnzoMethodMHDVlct::compute`. An old benchmark I ran indicates that this should give a considerable boost in performance.
- I'm interested in playing around with the implementation in `EnzoEFltMap` to try to allow better optimization of the Riemann solvers. One could imagine adding an extra level of indirection into `EnzoEFltMap` and storing the entries in an instance of `Cello<enzo_float,4>` (with the ordering of quantities optimized for the Riemann solver).


src/Enzo/enzo_EnzoConstrainedTransport.cpp
```diff
@@ -10,30 +10,289 @@
...
+	  ("This should not be called more than %d times to transition "
+	   "between the %d partial timestep(s)"),
+	  num_partial_timesteps_-1, num_partial_timesteps_,
+	  num_partial_timesteps_ > partial_timestep_index_);
+}
+
+//----------------------------------------------------------------------
+
+void EnzoConstrainedTransport::correct_reconstructed_bfield
+(EnzoEFltArrayMap &l_map, EnzoEFltArrayMap &r_map, int dim, int stale_depth)
```
pgrete, line 129:
```suggestion
(const EnzoEFltArrayMap &l_map, const EnzoEFltArrayMap &r_map, int dim, int stale_depth)
```

src/Enzo/enzo_EnzoConstrainedTransport.cpp
```diff
@@ -10,30 +10,289 @@
...
+      }
+    }
+    // Equivalently: l_bfield.subarray() = bfield;
+    //               r_bfield.subarray() = bfield;
+  }
+}
+
+//----------------------------------------------------------------------
+
+void EnzoConstrainedTransport::identify_upwind(EnzoEFltArrayMap &flux_map,
```
pgrete, line 174:
```suggestion
void EnzoConstrainedTransport::identify_upwind(const EnzoEFltArrayMap &flux_map,
```

src/Enzo/enzo_EnzoConstrainedTransport.cpp
```diff
@@ -10,30 +10,289 @@
...
+  // First, compute the edge-centered Electric fields (each time, it uses
+  // the current integrable quantities)
+  EnzoConstrainedTransport::compute_all_edge_efields(cur_prim_map, xflux_map,
+                                                     yflux_map, zflux_map,
+                                                     center_efield_,
+                                                     edge_efield_l_, weight_l_,
+                                                     stale_depth);
+
+  EnzoBlock *enzo_block = enzo::block(block_);
+  std::array<enzo_float,3> cell_widths;
```
pgrete, line 252: Why is this temp variable needed (versus directly using ` enzo_block->CellWidth`)?

src/Enzo/enzo_EnzoConstrainedTransport.cpp
```diff
@@ -10,30 +10,289 @@
...
+
+  // Update longitudinal B-field (add source terms of constrained transport)
+  for (int dim = 0; dim<3; dim++){
+    EnzoConstrainedTransport::update_bfield
+      (cell_widths, dim, edge_efield_l_, (*cur_bfieldi_l)[dim],
+       (*out_bfieldi_l)[dim], dt, stale_depth);
+  }
+
+  // Finally, update cell-centered B-field
+  const char* names[3] = {"bfield_x", "bfield_y", "bfield_z"};
```
pgrete, line 265: std::string?

src/Enzo/enzo_EnzoConstrainedTransport.cpp
```diff
@@ -10,30 +10,289 @@
...
+       stale_depth);
+  }
+}
+
+//----------------------------------------------------------------------
+
 void EnzoConstrainedTransport::compute_center_efield
-(Block *block, int dim, std::string center_efield_name, Grouping &prim_group,
- int stale_depth)
+(int dim, EFlt3DArray &efield, EnzoEFltArrayMap &prim_map, int stale_depth)
```
pgrete, line 276:
```suggestion
(int dim, EFlt3DArray &efield, const EnzoEFltArrayMap &prim_map, int stale_depth)
```

pgrete (commented):
This review is still WIP. I'll continue tomorrow.

From a style point of view I started to add many comments to make parameters const reference rather than just references (e.g., all the maps), and also started to mark `const char *` -> `std::string`.
Given that I'm not sure if you think it's worth updating I'll not add additional comments at those places.

src/Enzo/enzo_EnzoMethodMHDVlct.cpp
```diff
@@ -641,419 +552,155 @@ void EnzoMethodMHDVlct::compute_flux_
...
-  // All 3 array objects are the same shape
-  for (int iz=0; iz<bfield.shape(0); iz++) {
-    for (int iy=0; iy<bfield.shape(1); iy++) {
-      for (int ix=0; ix<bfield.shape(2); ix++) {
-	l_bfield(iz,iy,ix) = bfield(iz,iy,ix);
-	r_bfield(iz,iy,ix) = bfield(iz,iy,ix);
-      }
-    }
+  // interfaces (should potentially be handled internally by reconstructor)
+  if (ct_handler != NULL) {
```
pgrete, line 556:
```suggestion
  if (ct_handler != nullptr) {
```

src/Enzo/enzo_EnzoMethodMHDVlct.cpp
```diff
@@ -641,419 +552,155 @@ void EnzoMethodMHDVlct::compute_flux_
...
-{
-  // allocate the temporary fields in grouping
-  for (std::size_t i = 0; i < group_names.size(); i++){
-    std::string group_name = group_names[i];
-    int num_fields = grouping.size(group_name);
-    for (int j = 0; j < num_fields; j++){
-      prep_reused_temp_field_(field, grouping.item(group_name,j), cx, cy, cz);
-    }
+  // Finally, have the ct handler record the upwind direction
+  if (ct_handler != NULL){
```
pgrete, line 597:
```suggestion
  if (ct_handler != nullptr){
```

src/Enzo/enzo_EnzoMethodMHDVlct.hpp
```diff
@@ -6,11 +6,15 @@
 /// @brief    [\ref Enzo] Declaration of the VL + CT (van Leer combined with
 ///           constrained transport) MHD method
 ///
-/// This class makes relies on the following component classses
+/// This class encapsulates relies on the following component classses
```
pgrete, line 9: This sentence is missing sth, doesn't it?

src/Enzo/enzo_EnzoMethodMHDVlct.hpp
```diff
@@ -19,71 +23,54 @@
 ///      Nearly every field overlaps between the two cases. Examples of the
 ///      primitives for adiabatic, ideal gas (without dual energy formalism):
 ///        - density         (both integrable and reconstructable)
-///        - velocity       (both integrable and reconstructable)
+///        - velocity        (both integrable and reconstructable)
 ///        - pressure        (just reconstructable)
 ///        - total_energy    (just integrable)
+///      When supportin magnetic fields, there is also:
```
pgrete, line 29:
```suggestion
///      When supporting magnetic fields, there is also:
```

src/Enzo/enzo_EnzoMethodMHDVlct.hpp
```diff
@@ -94,6 +81,20 @@ class EnzoMethodMHDVlct : public Method {
   /// @ingroup  Enzo
   /// @brief    [\ref Enzo] Encapsulate VL + CT MHD method
 
+public:
+  /// This is defined within the scope of EnzoMethodMHDVlct to avoid polluting
+  /// the global namespace.
+  ///
+  /// @note
+  /// This must be public so that it can be passed to helper methods by value
+  enum bfield_choice {
```
pgrete, line 90: Could probably be made an `enum class`

src/Enzo/enzo_EnzoNestedPassiveScalarFieldList.cpp
```diff
@@ -0,0 +1,63 @@
...
+#include "cello.hpp"
+#include "enzo.hpp"
+#include "charm_enzo.hpp"
+
+// define a shorthand alias type just for use in this f
+typedef std::vector<std::string> str_vec_t;
+
+//----------------------------------------------------------------------
+
+std::shared_ptr<const std::vector<std::vector<std::string>>>
```
pgrete, line 18:
```suggestion
std::shared_ptr<const std::vector<str_vec_t>>
```

src/Enzo/enzo_EnzoNestedPassiveScalarFieldList.hpp
```diff
@@ -0,0 +1,86 @@
...
+      nested_names_ = build_nested_list_();
+      initialized_ = true;
+    }
+    return nested_names_;
+  }
+
+  /// Retrieve the nested passive scalar field list
+  ///
+  /// This method is required to support the timestep method of Method objects.
+  std::shared_ptr<const std::vector<std::vector<std::string>>> get_list()
```
pgrete, line 68:
```suggestion
  std::shared_ptr<const std::vector<str_vec_t>> get_list()
```

src/Enzo/enzo_EnzoReconstructor.cpp
```diff
@@ -12,31 +12,46 @@
...
   ASSERT("EnzoReconstructor::construct_reconstructor",
 	 "theta_limiter must satisfy 1<=theta_limiter<=2",
 	 (1.<=theta_limiter) && (theta_limiter<=2));
 
-  // some repeated code from construct_riemann
-  std::vector<std::string> groups = reconstructable_groups;
-  groups.insert(groups.end(), passive_groups.begin(), passive_groups.end());
+  // Construct a vector of keys for all components of the quantities listed in
+  // active_reconstructed_quantities
+  std::vector<std::string> key_l;
```
pgrete, line 25:
```suggestion
  str_vec_t key_l;
```

src/Enzo/enzo_EnzoReconstructor.hpp
```diff
@@ -59,32 +55,40 @@ class EnzoReconstructor : public PUP::able
...
+  /// shapes of cell-centered arrays (i.e. the same shape as arrays in
+  /// `prim_map`). In this case, the function effectively treats such arrays as
+  /// if their `subarray` method were invoked, where `CSlice(0,-1)` is
+  /// specified for the `dim` axis and `CSlice(nullptr,nullptr)` is specified
+  /// for other axes.
+  virtual void reconstruct_interface
+  (EnzoEFltArrayMap &prim_map, EnzoEFltArrayMap &priml_map,
+   EnzoEFltArrayMap &primr_map, int dim, EnzoEquationOfState *eos,
+   int stale_depth,
+   const std::vector<std::vector<std::string>>& passive_lists)=0;
```
pgrete, line 91: I was just about to suggest replacing `std::vector<std::string>` with `str_vec_t` again and am now wondering if it'd be worth to introduce a `passive_list_t` of `std::vector<std::vector<std::string>>`. What do you think?

src/Enzo/enzo_EnzoRiemannImpl.hpp
```diff
@@ -523,43 +460,46 @@ inline void passive_advection_helper_
...
-
-  for (std::string group_name : passive_scalar_groups_){
+  // First address passive scalars:
+  // passive_lists holds different lists of passive scalar quantities
+  // - The first entry in passive_lists (even if it's empty) provides a set of
+  //   keys for quantities that are simply passively advected
+  // - Later entries specify keys for groups of passive scalars that need to
+  //   sum to 1 (when in specific form).
+
+  bool first_list = true;
```
pgrete, line 475: Is this var every used?

pgrete (approved):
Nice work @mabruzzo 
I think this clarifies the implementation.
Effectively all comments I left are minor.

src/Cello/array_CelloArray.hpp
```diff
@@ -894,6 +911,24 @@ class CelloArray : public FixedDimArray_<T,D>
...
+                     this->shape_, this->stride_,
+                     (void*)other.shared_data_.get(), other.offset_,
+                     other.shape_, other.stride_, D);
+  }
+  bool is_alias(const TempArray_<T,D>& other) const{
+    return is_alias_((void*)this->shared_data_.get(), this->offset_,
+                     this->shape_, this->stride_,
+                     (void*)other.shared_data_.get(), other.offset_,
+                     other.shape_, other.stride_, D);
+  }
```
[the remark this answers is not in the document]
mabruzzo: I don't think so. I think I was just being lazy with figuring out how to properly template on the container type.

I'll address this

src/Enzo/enzo_EnzoIntegrableUpdate.hpp
```diff
@@ -40,79 +39,91 @@ class EnzoIntegrableUpdate : public PUP::able
...
   ///     computed.
-  void accumulate_flux_component(Block *block, int dim, double dt,
-				 Grouping &flux_group, Grouping &dUcons_group,
-				 int stale_depth) const;
+  /// @param[in]     passive_lists A list of lists of keys for passively
+  ///     advected scalars. Keys must not be duplicated across more than one
+  ///     list. In this method, this is effectively concatenated into one list
+  ///     of passive scalar keys.
+  void accumulate_flux_component
+  (int dim, double dt, enzo_float cell_width, EnzoEFltArrayMap &flux_map,
```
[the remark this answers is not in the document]
mabruzzo: The `timestep` virtual method of the `Method` class always computes the timestep as a double precision value (regardless of the precision of `enzo_float`). I'm somewhat inclined to keep this as a `double` to stay with this convention. But I could probably be convinced to change this.

src/Enzo/enzo_EnzoConstrainedTransport.cpp
```diff
@@ -10,30 +10,289 @@
...
+  ASSERT("EnzoConstrainedTransport", "num_partial_timesteps must be positive",
+	 num_partial_timesteps_ > 0);
+
+  if (num_partial_timesteps_ != 2){
+    ERROR("EnzoConstrainedTransport",
+          "This machinery hasn't been tested for cases when "
+          "num_partial_timesteps!=2.");
+  }
+
+  const char* field_names[] = {"bfieldi_x", "bfieldi_y", "bfieldi_z"};
```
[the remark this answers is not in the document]
mabruzzo: No not really. I think I was trying to be a little clever with string literals, (I vaguely recall reading something about `std::string` not representing string literals in c++11 - but I could be wrong), or making sure they were stack-allocated.

Either way, this is a micro-optimization and we would be probably be better of with using `std::string`. I'll change this.

src/Enzo/enzo_EnzoConstrainedTransport.cpp
mabruzzo: I'll fix this

src/Enzo/enzo_EnzoConstrainedTransport.cpp
mabruzzo: I think I originally was holding off on this because I wasn't sure about the desire behavior for a `const EnzoEFltArrayMap`... Do we want it to hold:
- `const CelloArray<enzo_float, 3>` where values within the array are still mutable
- `const CelloArray<const enzo_float, 3>`. Strictly speaking, this isn't supported by CelloArray, but I've adopted the Kokkos documentation. The idea is that the contents are read-only (going with this approach would require that we add this feature to the CelloArray.

I'm still not sure which approach we want to go with. I would appreciate feedback on this point.

You've pointed out other places where this distinction is moot and I should make the references const, either way.

src/Enzo/enzo_EnzoEOSIdeal.cpp
```diff
@@ -203,52 +251,64 @@ void EnzoEOSIdeal::pressure_from_integrable(Block *block,
   }
 
   const bool idual = this->uses_dual_energy_formalism();
+  const bool mag = (integrable_map.contains("bfield_x") ||
+                    integrable_map.contains("bfield_y") ||
+                    integrable_map.contains("bfield_z"));
```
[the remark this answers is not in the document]
mabruzzo: Done

src/Enzo/enzo_EnzoMethodMHDVlct.hpp
mabruzzo: As I recall, there are some weird issues with `PUP`ing instances of `enum class`. In particular I think I encountered an issue on Frontera or Stampede2 while using the intel compiler.

src/Enzo/enzo_EnzoMethodMHDVlct.cpp
```diff
@@ -184,214 +225,121 @@ std::vector<std::string> unique_combination_(const std::vector<std::string> &a,
...
-    // conserved-form of passively advected scalars (the initial values are
-    // taken from here and converted to specific form. Intermediate and final
-    // values get written to here)
-    Grouping *conserved_passive_scalars = cello::field_descr()->groups();
-
-    // allocate constrained transport object
-    EnzoConstrainedTransport ct = EnzoConstrainedTransport();
+    const enzo_float* const cell_widths = enzo::block(block)->CellWidth;
 
     double dt = block->dt();
```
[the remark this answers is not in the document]
mabruzzo: This bothers me a lot too. I think that this is also an issue with `cell_widths`. I'll open an issue to address this

src/Enzo/enzo_EnzoRiemannImpl.hpp
mabruzzo: Good catch

src/Enzo/enzo_EnzoMethodMHDVlct.cpp
```diff
@@ -113,52 +169,37 @@ void EnzoMethodMHDVlct::determine_quantities_
...
-void add_passive_groups_(Grouping &grouping, const std::string field_prefix,
-			 const std::vector<std::string> group_names)
+EnzoMethodMHDVlct::~EnzoMethodMHDVlct()
 {
-  // Helper function that adds entries of passively advected groups (specified
-  // in the configuration file) to an existing Grouping object
+  delete eos_;
+  delete half_dt_recon_;
+  delete full_dt_recon_;
+  delete riemann_solver_;
```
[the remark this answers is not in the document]
mabruzzo: good catch

mabruzzo:
I finished updating documentation and I hope to finish addressing the comments this week

src/Enzo/enzo_EnzoIntegrableUpdate.hpp
```diff
@@ -40,79 +39,91 @@ class EnzoIntegrableUpdate : public PUP::able
...
+  /// @param[in]  eos Pointer to the fluid's equation of state object. When
+  ///     applicable used for placing a density floor.
+  /// @param[in]  stale_depth The stale depth at the time of the function call
+  ///     (the stale_depth must be incremented after this function is called)
+  /// @param[in]  passive_lists A list of lists of keys for passively advected
+  ///     scalars. The first list holds the keys for quantities that are
+  ///     normally passively advected. Subsequent lists group together
+  ///     collections pf passively advected scalars whose specific value
+  ///     (before and after advection) must sum to 1. Keys must not be
+  ///     duplicated across more than one list.
```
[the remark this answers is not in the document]
mabruzzo: > I'm asking because if this is a "sometime in the future" feature that may or may not be implemented by you, then it may be cleaner to remove that implicit logic and just go with a single list for now (and then make the update when additional lists will actually be supported).

Good point. I'll look into this. I think most of the pieces should be in place. It may be easier just to add this feature than to remove the logic.

src/Enzo/enzo_EnzoConstrainedTransport.hpp
```diff
@@ -12,166 +12,304 @@ class EnzoConstrainedTransport
...
+  /// The instance also internally tracks the current partial timestep to be
+  /// less intrusive for the underlying MHD integrator.
+  ///
+  /// Although this design is crude, it is simple. The design could be improved
+  /// by making separating the responsiblities of the allocation/deallocation
+  /// of temporary scratch space and the implementation of the constrained
+  /// transport operation between two classes. However this refactoring has
+  /// been defered to the future when other classes may be introduced to
+  /// encapsulate other approaches to updated the magnetic field (e.g. Dedner
+  /// divergence cleaning).
```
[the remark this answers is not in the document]
mabruzzo: You're absolutely right. I've started to give this some thought and the thing that I'm really concerned about is reentrancy. For example if somebody plopped a Charm++ reduction into the middle of `EnzoMethodMHDVlct::compute` (for example, to facillitate Riemann Solver fallback), this could break Constrained Transport in unexpected ways if we aren't careful.

In reality, some comments about this in the code is probably sufficient.

mabruzzo:
Just a minor update: I'm making some final tweaks today that should be finished today (largely in response to some review comments). Then, this should be ready for final review.

mabruzzo:
@pgrete (and @forrestglines) Apologies that this took so long to wrap up. I was basically done yesterday morning, but I kind of went down a rabbit hole. I think you're good to take one more look and we'll (hopefully) be done.

**Passive Scalar List**
First I completely removed the nested list of list pattern used to specify the names of passive scalars. I talked it over with Greg and we don't think it's necessary for the MHD integrator to handle (if it turns out to be important we can always add it in later). I updated the developer documentation and excised some unused code from the Riemann Solver that had been written a while back in anticipation of supporting this feature. I also renamed the ``EnzoNestedPassiveScalarFieldList`` class as ``EnzoLazyPassiveScalarFieldList``.

 
**Constrained Transport Refactor**
Second, in response to some of your comments about the CT implementation, I started to refactor it so that a single instance could be reused for multiple blocks and cycles. To do this, I kind of embraced the fact the my implementation was acting like a State Machine (i.e. I started to emphasize this in my description so that it's easier for other people to understand what is going on). After completing a fair amount of refactoring, I was finding it hard to disentangle the state machine logic from the actual Bfield integration code.

To try to disentangle it, I decided to create the ``EnzoBfieldMethod`` abstract base class that basically juss implements the state machine and refactored ``EnzoConstrainedTransport`` so that it was a subclass that basically just implemented the integration steps (I changed the name from ``EnzoConstrainedTransport`` to ``EnzoBfieldMethodCT``).  In theory, ``EnzoBfieldMethod`` might have subclasses to implement different magnetic field integration strategies.  As you may recall, this basic designed was briefly discussed during PR #62. This also conforms to the overall design of the other Hydro tools.

Unfortunately this was a much bigger undertaking than I had initially anticipated and required a lot more boilerplate code than I had expected. Alongside this refactor I also updated the website's developer documentation to describe how to use ``EnzoBfieldMethod``.

I just wanted to note:
- The name may need to be improved (it might be a little confusing since the presence of "Method" in a class name usually indicates that it has been derived from cello's ``Method`` class). Some other names might include ``EnzoBfieldAlgo`` or ``EnzoBfieldStrategy``. I'm curious to know if you have other ideas/preferences.
- I'm not entirely convinced that we should be using subclasses here since the choice of magnetic field integration seems so fundamental to the integrator, but I think it's probably fine since we are performing these operations at the block level.

Given the size of this change, would you prefer that I drop the last 2 commits from this PR that introduce this change and submit them in a new, separate PR?

@forrestglines, I'm not sure whether you've started implementing B-field flux corrections. If you have and this change creates issues, I would be very happy to help address them.

**Outstanding item**
I think there are 2 outstanding items from @pgrete's review is potentially dropping ``integrable_to_reconstructable`` and ``reconstructable_to_integrable`` from ``EnzoEquationOfState``.

I plan to address that in the PR where that I mentioned over slack (where I'm going to rename the variables and drop the aliasing). In that PR, I'll assess whether a revised version of that method is helpful.

mabruzzo:
Note, that the most recent commit was just accounting for a merge conflict that had been introduced in the website documentation. Test-suite-2 definitely passes

src/Enzo/enzo_EnzoRiemannImpl.hpp
```diff
@@ -306,40 +291,18 @@ void EnzoRiemannImpl<ImplFunctor>::pup (PUP::er &p)
...
 void EnzoRiemannImpl<ImplFunctor>::solve
-(Block *block, Grouping &priml_group, Grouping &primr_group,
- std::string pressure_name_l, std::string pressure_name_r,
- Grouping &flux_group, int dim, EnzoEquationOfState *eos, int stale_depth,
- std::string interface_velocity_name) const
+(EnzoEFltArrayMap &prim_map_l, EnzoEFltArrayMap &prim_map_r,
+ const EFlt3DArray &pressure_array_l, const EFlt3DArray &pressure_array_r,
+ EnzoEFltArrayMap &flux_map, int dim, EnzoEquationOfState *eos,
+ int stale_depth, const str_vec_t &passive_list,
+ EFlt3DArray *interface_velocity) const
```
forrestglines, line 305:
```suggestion
(EnzoEFltArrayMap &prim_map_l, EnzoEFltArrayMap &prim_map_r,
 const EFlt3DArray &pressure_array_l, const EFlt3DArray &pressure_array_r,
 EnzoEFltArrayMap &flux_map, const int &dim, EnzoEquationOfState *eos,
 const int &stale_depth, const str_vec_t &passive_list,
 EFlt3DArray *interface_velocity) const
```

src/Enzo/enzo_EnzoRiemannImpl.hpp
```diff
@@ -428,92 +375,55 @@ void EnzoRiemannImpl<ImplFunctor>::solve
...
 //----------------------------------------------------------------------
 
 inline void passive_advection_helper_
-(std::string group_name, Grouping &priml_group, Grouping &primr_group,
- Grouping &flux_group, const EFlt3DArray &density_flux, int dim,
- EnzoFieldArrayFactory array_factory, const bool unity_sum) noexcept
+(const str_vec_t &passive_keys,
+ EnzoEFltArrayMap& prim_map_l, EnzoEFltArrayMap& prim_map_r,
+ EnzoEFltArrayMap& flux_map, const EFlt3DArray &density_flux,
+ int stale_depth) noexcept
```
forrestglines, line 396:
```suggestion
(const str_vec_t &passive_keys,
 EnzoEFltArrayMap& prim_map_l, EnzoEFltArrayMap& prim_map_r,
 EnzoEFltArrayMap& flux_map, const EFlt3DArray &density_flux,
 const int &stale_depth) noexcept
```

src/Enzo/enzo_EnzoRiemannImpl.hpp
```diff
@@ -428,92 +375,55 @@ void EnzoRiemannImpl<ImplFunctor>::solve
...
 {
-  int num_fields = priml_group.size(group_name);
-  if (num_fields == 0) {return;}
-  EnzoPermutedCoordinates coord(dim);
+  std::size_t num_keys = passive_keys.size();
```
forrestglines, line 398:
```suggestion
  const std::size_t num_keys = passive_keys.size();
```

src/Enzo/enzo_EnzoRiemannImpl.hpp
```diff
@@ -428,92 +375,55 @@ void EnzoRiemannImpl<ImplFunctor>::solve
...
+  const int sd = stale_depth;
+  for (int iz = sd; iz < density_flux.shape(0) - sd; iz++) {
+    for (int iy = sd; iy < density_flux.shape(1) - sd; iy++) {
+
+      for (int key_ind=0; key_ind<num_keys; key_ind++){
+        for (int ix = sd; ix < density_flux.shape(2) - sd; ix++) {
+          enzo_float dens_flux = density_flux(iz,iy,ix);
+          enzo_float wl = wl_arrays[key_ind](iz,iy,ix);
+          enzo_float wr = wr_arrays[key_ind](iz,iy,ix);
+          enzo_float w_upwind = (dens_flux>0) ? wl : wr;
```
forrestglines, line 422:
```suggestion
          const enzo_float dens_flux = density_flux(iz,iy,ix);
          const enzo_float wl = wl_arrays[key_ind](iz,iy,ix);
          const enzo_float wr = wr_arrays[key_ind](iz,iy,ix);
          const enzo_float w_upwind = (dens_flux>0) ? wl : wr;
```

src/Enzo/enzo_EnzoRiemannImpl.hpp
```diff
@@ -428,92 +375,55 @@ void EnzoRiemannImpl<ImplFunctor>::solve
...
-							     field_ind, dim);
-    wr_arrays[field_ind] = array_factory.reconstructed_field(primr_group,
-							     group_name,
-							     field_ind, dim);
-    flux_arrays[field_ind] = array_factory.from_grouping(flux_group,
-							 group_name,
-							 field_ind);
+  EFlt3DArray *wl_arrays = new EFlt3DArray[num_keys];
+  EFlt3DArray *wr_arrays = new EFlt3DArray[num_keys];
+  EFlt3DArray *flux_arrays = new EFlt3DArray[num_keys];
```
forrestglines, line 406:
```suggestion
  const EFlt3DArray *wl_arrays = new EFlt3DArray[num_keys];
  const EFlt3DArray *wr_arrays = new EFlt3DArray[num_keys];
  const EFlt3DArray *flux_arrays = new EFlt3DArray[num_keys];
```
This might work - in which case would be more robust.

src/Enzo/enzo_EnzoRiemannImpl.hpp
```diff
@@ -523,43 +433,31 @@ inline void passive_advection_helper_
 
 template <class ImplFunctor>
 void EnzoRiemannImpl<ImplFunctor>::solve_passive_advection_
-(Block* block, Grouping &priml_group, Grouping &primr_group,
- Grouping &flux_group, const EFlt3DArray &density_flux, int dim,
- int stale_depth)
-  const throw()
+(EnzoEFltArrayMap &prim_map_l, EnzoEFltArrayMap &prim_map_r,
+ EnzoEFltArrayMap &flux_map, const EFlt3DArray &density_flux,
+ int stale_depth, const str_vec_t &passive_list) const throw()
```
forrestglines, line 438:
```suggestion
(EnzoEFltArrayMap &prim_map_l, EnzoEFltArrayMap &prim_map_r,
 EnzoEFltArrayMap &flux_map, const EFlt3DArray &density_flux,
 const int &stale_depth, const str_vec_t &passive_list) const throw()
```

src/Enzo/enzo_EnzoRiemannUtils.hpp
```diff
@@ -166,89 +166,60 @@ namespace enzo_riemann_utils{
...
   ///   respectively. Note that each of the fields in grouping are assumed to
   ///   be face-centered along this dimension (excluding the exterior faces of
   ///   the mesh). This allows for appropriate loading of reconstructed fields.
-  /// @param stale_depth indicates the current stale_depth for the loaded
-  ///   quanties.
   template<class LUT>
   inline std::array<EFlt3DArray,LUT::NEQ> load_array_of_fields
-  (Block *block, Grouping &grouping,
-   int dim, int stale_depth) noexcept
+  (EnzoEFltArrayMap& map, int dim) noexcept
```
forrestglines, line 210:
```suggestion
  (EnzoEFltArrayMap& map, const int &dim) noexcept
```

src/Enzo/enzo_EnzoRiemannUtils.hpp
```diff
@@ -166,89 +166,60 @@ namespace enzo_riemann_utils{
...
     EnzoPermutedCoordinates coord( (dim == -1) ? 0 : dim);
-    EnzoFieldArrayFactory array_factory(block, stale_depth);
 
-    // define a lambda function to execute for every member of lut
-    auto fn = [coord, dim, &arr, &array_factory, &grouping](std::string name,
-                                                            int index)
+    // define a lambda function to execute for every member of lut. For each
+    // member in lut, its passed: 1. the member's name
+    //                            2. the associated index
+    auto fn = [coord, &arr, &map](std::string name, int index)
```
forrestglines, line 220:
```suggestion
    auto fn = [coord, &arr, &map](const std::string &name, const int &index)
```

src/Enzo/enzo_EnzoSourceInternalEnergy.cpp
```diff
@@ -11,9 +11,9 @@
 //----------------------------------------------------------------------
 
 void EnzoSourceInternalEnergy::calculate_source
-(Block *block, double dt, Grouping &prim_group, Grouping &dUcons_group,
- std::string interface_velocity_name, int dim, EnzoEquationOfState *eos,
- int stale_depth) const throw()
+(int dim, double dt, enzo_float cell_width, EnzoEFltArrayMap &prim_map,
+ EnzoEFltArrayMap &dUcons_map, EFlt3DArray &interface_velocity,
+ EnzoEquationOfState *eos, int stale_depth) const throw()
```
forrestglines, line 16:
```suggestion
(const int &dim, const double &dt, const enzo_float &cell_width, EnzoEFltArrayMap &prim_map,
 EnzoEFltArrayMap &dUcons_map, EFlt3DArray &interface_velocity,
 EnzoEquationOfState *eos, const int &stale_depth) const throw()
```

src/Enzo/enzo_EnzoSourceInternalEnergy.cpp
```diff
@@ -24,10 +24,8 @@ void EnzoSourceInternalEnergy::calculate_source
 	 "The EOS can't be barotropic and use the dual energy formalism.",
 	 !(eos->is_barotropic()) );
 
-  EnzoBlock * enzo_block = enzo::block(block);
-  enzo_float dtdx = dt/enzo_block->CellWidth[dim];
+  enzo_float dtdx = dt/cell_width;
```
forrestglines, line 27:
```suggestion
  const enzo_float dtdx = dt/cell_width;
```

src/Enzo/enzo_EnzoSourceInternalEnergy.hpp
```diff
@@ -51,40 +51,38 @@ class EnzoSourceInternalEnergy
...
+  /// @param[in]  eos Instance of the fluid's EnzoEquationOfState object
+  /// @param[in]  stale_depth indicates the current stale_depth for the
+  ///     supplied cell-centered quantities. The update using the
+  ///     reconstructor's delayed_staling_rate should be applied at some
+  ///     time after this function call.
+  void calculate_source(int dim, double dt, enzo_float cell_width,
+                        EnzoEFltArrayMap &prim_map,
+                        EnzoEFltArrayMap &dUcons_map,
+                        EFlt3DArray &interface_velocity,
+                        EnzoEquationOfState *eos, int stale_depth)
```
forrestglines, line 85:
```suggestion
  void calculate_source(const int &dim, const double &dt, const enzo_float &cell_width,
                        EnzoEFltArrayMap &prim_map,
                        EnzoEFltArrayMap &dUcons_map,
                        EFlt3DArray &interface_velocity,
                        EnzoEquationOfState *eos, const int &stale_depth)
```

forrestglines (approved):
Just finished my review. Very good work @mabruzzo 

I made a few changes with regards to using `const &` types in a few places. Those changes aren't necessary, but if they compile I think they would be more robust.

The CI test isn't passing - but it looks like the test is timing out? Do you know why that was  encountered with the last commit?

src/Enzo/enzo_EnzoRiemannImpl.hpp
mabruzzo: Yes, it should definitely work, since `CelloArray` has pointer semantics. 

The shape and the pointer to array memory of a `const CelloArray<T, D>` can't be changed, but the contents of the array still can be changed.

As an aside, I was inspired by Kokkos and I'm currently working on a PR to support `CelloArray<const T, D>`, which is equivalent to a pointer to a constant (I'm adding support for implicit casting from `CelloArray<T, D>` to `CelloArray<const T, D>` and making sure all of the methods behave properly).

src/Enzo/enzo_EnzoBfieldMethod.hpp
```diff
@@ -0,0 +1,218 @@
...
+  }
+
+  int num_partial_timesteps() const noexcept {return num_partial_timesteps_;}
+
+private: // keep subclasses from touching these attributes
+
+  /// Number of partial timesteps in a cycle. A value of 1 would means that
+  /// there is only a single update over the full timestep. A value of 2 means
+  /// that there is a half timestep and a full timestep.
+  int num_partial_timesteps_;
```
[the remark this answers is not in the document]
mabruzzo: Currently the integration scheme is just 2nd order accurate in time (Van Leer).

To answer your question: Yes, it was my intention to make the implementation flexible enough, that it could be easily extended to other integration orders. I think it should be trivial to modify `EnzoMethodMHDVlct` so that you could switch between the Van Leer scheme and a 3rd order integration in time scheme (we would need to change the name of the class in that case). But in the worst case scenario, you could definitely reuse all of the MHD helper classes.

As far as I can tell, we would need to make 3 main tweaks:
1. We need to be able the number of times we enter the for-loop in VL+CT and tweak our handling of the temporary arrays used for holding values at partial timesteps
2. `EnzoBfieldMethod` would need to be updated to support the new number of partial timesteps
3. If using something like a Runge-Kutta scheme, then `EnzoIntegrableUpdate::update_quantities` and `EnzoBfieldMethod::update_all_bfield_components` need to be tweaked since the final fluid properties are based on an average of the properties at the partial timesteps.

I don't it's necessary, but if the ghost zones do need to be updated at partial timesteps that might take some extra work. I'm pretty sure it's possible; there has been talk of using this sort of feature for refreshing particles for stellar feedback and ray-tracing.

I can talk more about this if you want (Philipp also seemed very optimistic that this is possible).

src/Enzo/enzo_EnzoIntegrableUpdate.hpp
```diff
@@ -40,79 +39,80 @@ class EnzoIntegrableUpdate : public PUP::able
...
+  ///     to. If constrained transport is being used, this will not include
+  ///     arrays for the magnetic fields.
+  /// @param[in]     stale_depth The stale depth at the time of the function
+  ///     call. This should match the stale depth at the time the fluxes were
   ///     computed.
-  void accumulate_flux_component(Block *block, int dim, double dt,
-				 Grouping &flux_group, Grouping &dUcons_group,
-				 int stale_depth) const;
+  /// @param[in]     passive_list A list of keys for passive scalars.
+  void accumulate_flux_component(int dim, double dt, enzo_float cell_width,
```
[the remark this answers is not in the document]
mabruzzo: @pgrete also made a similar suggestion. The use of `double` is very intentional here because `Method` subclass instances **always** compute the timestep (via the virtual `Method::timestep (Block * block) const` method) in double precision. This is entirely independent of the precision of `enzo_float`.

Within the context of `EnzoMethodMHDVlct`, I've tried to be self-consistent about always storing the (partial) timestep in double precision. I think changing that is somewhat beyond the scope of this PR (if we want to change it, I think we should make sure that all `Method` objects handle it self-consistently).

mabruzzo:
@forrestglines Thank for looking this over! 

Would you mind briefly elaborating on why you think it would be better to pass primitive arguments (they have types of `int`, `double`, `enzo_float`) by constant reference instead of by value? I was under the impression that it is actually slightly better to pass these small types by value (to avoid an unnecessary level of indirection). Is that wrong? (I might be over-generalizing something I read)

If there isn't any benefit, to passing the primitive types by const reference, I'm inclined to skip those particular suggested changes in order to minimize the number of conflicts that I introduce with PR #94. Most of the other recommendations seem really useful!

mabruzzo:
The test failure when compiled in double-precision mode (i.e. test-suite-2) is due to one of the (non-VLCT) test problems taking too long to run. Sometimes you can get lucky and the test problem will complete running before the timeout occurs (maybe it depends on what else is running on the CI server at the same time...). A similar thing happens with test-suite-1 (compilation in single-precision mode), but the test problem in question usually completes in time after restarting the test suite once or twice.  James said that he has introduced some changes in his branch that ensure that this particular test takes less time to run to completion.

I'm extremely confident that this PR currently passes the tests (the test passed before the commit from 3 days ago which just altered some documentation links). I'll need to be more certain after accepting your recommended changes.

forrestglines:
> Would you mind briefly elaborating on why you think it would be better to pass primitive arguments (they have types of `int`, `double`, `enzo_float`) by constant reference instead of by value? I was under the impression that it is actually slightly better to pass these small types by value (to avoid an unnecessary level of indirection). Is that wrong? (I might be over-generalizing something I read)

I think you're right about reference types, but I think including `const` would be more correct.

> If there isn't any benefit, to passing the primitive types by const reference, I'm inclined to skip those particular suggested changes in order to minimize the number of conflicts that I introduce with PR #94. Most of the other recommendations seem really useful!

That's fine to defer these corrections for later.

pgrete (approved):
I just re-reviewed and didn't spot anything major yet we didn't discuss.
Great work! I'll pull the trigger.

PR #82, opened 2021-02-11
Add machine file and update doc for node local runs

pgrete:
Assuming we want additional machine file also for smaller centers, this one uses only module provided by the center (i.e., apart from grackle no custom libraries need to be installed).
I also added a note on using `++local` as this came up now several times on Slack.


mabruzzo (approved):

forrestglines (approved):

jwise77 (approved):

pgrete:
Got three approvals. Pulling the trigger.

PR #84, opened 2021-02-25
Update SCons to 4.1.0

pgrete:
The SCons version currently shipped with Enzo-E required Python2, which is EOL.
The newer version now supports Python3 (>=3.5).


pgrete:
NB: depending how people feel about git submodules, we likely also ship Enzo-E with SCons as git submodule so that we don't need to include the entire package.

mabruzzo:
I wanted to quickly point out that PR #46 tries to do something similar (you may already know this). It updates the SCons files so that we can use python3, but it depends on a user-installed version on SCons.

We might want to try to gather a consensus on which option people prefer before merging either PR. (I don't feel very strongly about - there are pros and cons with either option)

pgrete:
Silly me, I lost sight of the wood for the trees. I wanted a compiledb and needed a newer version of SCons so I just quickly updated it.
I'm also in favor of not shipping a version of SCons with the Enzo-E source and use a packaged version through #46

PR #88, opened 2021-04-14
Write New Outputs on Restart

mabruzzo:
In this PR I introduced the capability to specify new output file-sets during a restart from a checkpoint. To keep things manageable, the user is NOT allowed to mutate the properties of file-sets that were previously specified. I also updated the website documentation to reflect this new capability. This involved shuffling the order of some Cello operations, so I definitely would like @jobordner to eventually review this.

Note that I dropped the comment from the documentation stating that `Field:courant` could be updated during a restart because a comment at the top of  `initial/Checkpoint/restart.incl` explains that a change in where the courant factor is specified, had broken this feature.

An updated version of `initial/Checkpoint/restart.incl` that makes use of these features can be found [here](https://pastebin.com/9rratFB7).

I still need to introduce an automated test checking this feature before this is ready for review, but I need to work on something else more immediately.

I also wanted to receive some feedback on a related topic:
---

For those who don't know, simulation properties are currently only updated during a restart by a special restart-parameter file. The path to this file must be specified with the `Restart:file` parameter when initially launching the simulation for the very first time. 

Unfortunate, this means that if a user runs an expensive simulation and wants to use the feature introduced in this PR to generate new outputs after restarting from a checkpoint, they need to anticipate this scenario before they even launch the simulation. If they didn't anticipate this need (or just forgot to specify `Restart:file`), they must rerun the entire simulation up to that point where they want to restart.

The fact that charm++ doesn't seem to allow the user to pass new arguments to the program during a restart seems to be the source of the problem. I have 2 potential solutions to address this problem, and I want to get some feedback:
1. I wrote a separate branch where the user can specify the output file via an ``CELLO_RESTART_FILE``. However, I found it difficult to make sure that the variable was properly in the charm++ runtime when I ran it on my local machine. It only seemed to work if I ran the code with `++local` (and there were some weird issues with restarting `input/Checkpoint/checkpoint_ppm-8.in` while `++local`). I also tried to use the `++runscript` option to set the environment variable, but couldn't get it to work. Consequently, I have serious concerns about this approach to work (especially on clusters with multiple nodes). But maybe my concerns aren't warranted (If we pursue this approach I would need help).
2. My alternative idea is much simpler. When  `Restart:file` isn't specified, we could just look for the restart parameter file at some default path (maybe `./restart.in`?), and we continue with the restart even if the file can't be found. When  `Restart:file` is specified, we would use the existing behavior (raise an error if no restart parameter file can be found at the specified path). I have another separate branch where I implemented a variation on this idea and it seems to work well.

My inclination is to go with solution 2, but I wanted to check to see if anybody had any thoughts or preferences

### DIFF
--- a/doc/source/param/restart.incl
+++ b/doc/source/param/restart.incl
@@ -4,4 +4,19 @@
 :Default: :d:`""`
 :Scope:     :c:`Cello`
 
-:e:`This optional parameter is used to specify the name of a parameter file to read on restart.  Its purpose is to allow a simulation to be restarted with slightly different parameter values, e.g. with a smaller courant number.  Currently, very few parameters are supported in the restart parameter file, just` :p:`Field` : :p:`courant` :e:`and` :p:`Testing` : :p:`time_final`.
+:e:`This optional parameter is used to specify the name of a parameter
+file to read on restart.  Its purpose is to allow a simulation to be
+restarted with slightly different parameter values.  Currently, very
+few parameters are supported in the restart parameter file. These
+include:`
+
+  * Modifying the value of :p:`Testing` : :p:`time_final`
+
+  * Specifying new output file sets. These should be specified in an
+    :p:`Output` Group in a similar manner to how the file set would be
+    specified in a normal parameter file. The entries of :p:`Output` :
+    :p:`list` should NOT include any of the names of file sets
+    specified in the original input file (the entries in the restart
+    file's :p:`Output` : :p:`list` are effectively appended to the
+    list of entries from the original file). Note that the properties
+    for existing file sets can NOT be be modified.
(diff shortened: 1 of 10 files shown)


DeovratPrasad (approved):

bwoshea:
@jobordner just checking in - could you possibly review this PR some time this week?  Thank you!

mabruzzo:
This isn't ready for review. Something is unexpectedly breaking in this PR when I use the MPI backend. And at this point, I have no plans to come back to this in the near future (I managed to resolve the issue with my simulations without this feature)

 @bwoshea would you prefer that I close this?

bwoshea:
@mabruzzo Yes, I think that it's fine for you to close this.  Regarding your question about how to correctly do restarts with changed parameters, did you resolve this, or do you still want feedback?  If the latter, perhaps you could put that into an Issue and we can use that as a forum for discussion?

Issue #83, opened 2021-02-23
Representation of particle mass versus density


jwise77:
I'm opening up an issue to have a discussion on the particle masses. This was brought up in #49. The particles in the group `has_mass` will be included in the gravity solve. Currently, the mass field is defined to be density. This is how Enzo handled the particle masses to ease their mass deposition into grids for the gravity solve.

Do we keep this convention from Enzo?  Or do we rename the field to `particle_mass_density` or the like? Then we could have some derived field that contains their masses. I could be convinced either way.

matthewturk:
This would somewhat reduce the complexity on the yt side.

gregbryan:
I think we should break with Enzo and define mass as mass. It will just simplify things and make the code easier to read and modify.

stefanarridge:
I agree @gregbryan. At the moment, as far as I can tell, the particle 'masses' are kept the same when blocks are refined or coarsened, which is not consistent with treating 'mass' as density. I believe that is the reason for these lines of code in [src/Enzo/enzo_EnzoMethodPmDeposit.cpp](https://github.com/enzo-project/enzo-e/blob/master/src/Enzo/enzo_EnzoMethodPmDeposit.cpp)
```
// Scale mass by volume if particle value is mass instead of density

      int level = block->level();

      // Required for Cosmology ("mass" is mass)
      // Not for Collapse ("mass" is density)
      if (cosmology) {
        dens *= std::pow(2.0,rank*level);
      }
```
This PmDeposit method would be simplified a lot by simply depositing the particle mass to the grid before dividing by the cell volume. One thing that I would like to clear up is the treatment of comoving vs proper quantities. There are lots of statements all over the code looking something like:

```
if (cosmology){
x *= cosmo_a;
}
```
I have generally found it quite confusing to keep track of whether quantities are comoving or proper. In this case, do we need to divide by the comoving or proper cell volume. I also discussed this in #85 with respect to the gravity time step. It might need to be the subject of a separate Issue.

```

mabruzzo:
Resolved by PR #89

PR #49, opened 2020-06-12
Isolated Galaxy - Feedback

aemerick:
Large PR containing changes I've made to implement and isolated galaxy propblem type following that done in the Agora isolated galaxy problem type in Enzo. I'll list a high level summary of all the changes in this PR below. 

This is a heavy WIP and depends strongly on other PRs (including solver-dd).

PR #89, opened 2021-04-28
Particle masses

stefanarridge:
This changes `EnzoMethodPmDeposit` so that that all particles in the 'has_mass' group are taken into account when calculating the mass field, and the particles' mass attribute/constant is always interpreted as a quantity with dimensions of mass, rather than sometimes density. This fixes issue #83.

This may cause some issues with some of the Collapse tests, since those were written assuming that 'mass' is really density.

### DIFF
--- a/src/Cello/data_ParticleDescr.cpp
+++ b/src/Cello/data_ParticleDescr.cpp
@@ -253,6 +253,24 @@ int ParticleDescr::attribute_index (int it, std::string attribute_name) const
 
 //----------------------------------------------------------------------
 
+bool ParticleDescr::has_attribute(int it, std::string attribute_name) const
+{
+  ASSERT1("ParticleDescr::is_attribute",
+          "Trying to access unknown particle type '%s'",
+          type_name(it).c_str(),
+          (0 <= it && it < num_types()));
+
+  auto iter=attribute_index_[it].find(attribute_name);
+
+  int index = (iter != attribute_index_[it].end()) ? iter->second : -1;
+
+  static int count[CONFIG_NODE_SIZE] = {0};
+
+  return !((index == -1) && (count[cello::index_static()]++ < 10));
+}
+
+//----------------------------------------------------------------------
+
 std::string ParticleDescr::attribute_name (int it, int ia) const
 {
   ASSERT2("ParticleDescr::attribute_name",
@@ -389,6 +407,23 @@ int ParticleDescr::constant_offset (int it, int ic) const
             (0 <= ic && ic < num_constants(it))));
   return constant_offset_[it][ic]; 
 }
+
+//----------------------------------------------------------------------
+
+bool ParticleDescr::has_constant(int it, std::string constant_name) const
+{
+  ASSERT1("ParticleDescr::constant_index",
+          "Trying to access unknown particle type '%s'",
+          type_name(it).c_str(),
+          (0 <= it && it < num_types()));
+
+  auto iter=constant_index_[it].find(constant_name);
+
+  bool check = (iter != constant_index_[it].end()) ? true : false;
+
+  return check;
+}
+
 
 //----------------------------------------------------------------------
 // Bytes
--- a/src/Enzo/enzo_EnzoMethodPmDeposit.cpp
+++ b/src/Enzo/enzo_EnzoMethodPmDeposit.cpp
@@ -2,6 +2,7 @@
 
 /// @file     enzo_EnzoMethodPmDeposit.cpp
 /// @author   James Bordner (jobordner@ucsd.edu)
+/// @author   Stefan Arridge (stefan.arridge@gmail.com)
 /// @date     Fri Apr  2 17:05:23 PDT 2010
 /// @brief    Implements the EnzoMethodPmDeposit class
 ///
@@ -104,22 +105,46 @@ void EnzoMethodPmDeposit::compute ( Block * block) throw()
     block->lower(&xm,&ym,&zm);
     block->upper(&xp,&yp,&zp);
     block->cell_width(&hx,&hy,&hz);
-
-    // declare particle position arrays
-
-    //--------------------------------------------------
-    // Add "dark" particle density
-    //--------------------------------------------------
     
-    if (particle.type_exists ("dark")) {
-      const int it = particle.type_index ("dark");
 
-      const int ic_mass = particle.constant_index (it,"mass");
-
-      enzo_float dens = *((enzo_float *)(particle.constant_value (it,ic_mass)));
-
-      // check precisions match
+    // Get cosmological scale factors, if cosmology is turned on
+    enzo_float cosmo_a=1.0;
+    enzo_float cosmo_dadt=0.0;
+    EnzoPhysicsCosmology * cosmology = enzo::cosmology();
     
+    if (cosmology) {
+      cosmology->compute_expansion_factor(&cosmo_a,&cosmo_dadt,
+					  block->time() + alpha_*block->dt());
+    }
+
+    // Stefan: Not sure if cell widths need to be converted to a physical
+    // length
+    //if (rank >= 1) hx *= cosmo_a;
+    //if (rank >= 2) hy *= cosmo_a;
+    //if (rank >= 3) hz *= cosmo_a;
+
+    double inv_vol = 1.0;
+    if (rank >= 1) inv_vol /= hx;
+    if (rank >= 2) inv_vol /= hy;
+    if (rank >= 3) inv_vol /= hz;
+
+    // Stefan: Unsure of the meaning of this line. How is cosmological time
+    // defined in Enzo?
+    const double dt = alpha_ * block->dt() / cosmo_a;
+
+    // Get the number of particle types in the 'has_mass' group
+    ParticleDescr * particle_descr = cello::particle_descr();
+    Grouping * particle_groups = particle_descr->groups();
+    int num_mass = particle_groups->size("has_mass");
+
+    // Loop over all particles that have mass
+    for (int ipt = 0; ipt < num_mass; ipt++) {
+      const int it = particle.type_index(particle_groups->item("has_mass",ipt));
+
+      // Index for mass attribute / constant
+      int im = 0;
+      
+      // check correct precision for position
       int ia = particle.attribute_index(it,"x");
       int ba = particle.attribute_bytes(it,ia); // "bytes (actual)"
       int be = sizeof(enzo_float);                // "bytes (expected)"
@@ -132,40 +157,51 @@ void EnzoMethodPmDeposit::compute ( Block * block) throw()
                ((be == 4) ? "single" : ((be == 8) ? "double" : "quadruple")),
                (ba == be));
 
-      // Accumulate particle density using CIC
 
-      enzo_float cosmo_a=1.0;
-      enzo_float cosmo_dadt=0.0;
-      EnzoPhysicsCosmology * cosmology = enzo::cosmology();
-    
-      if (cosmology) {
-
-        double time = block->time();
-        double dt   = block->dt();
-        cosmology->compute_expansion_factor (&cosmo_a,&cosmo_dadt,time+alpha_*dt);
-      
-      }
-      if (rank >= 1) hx *= cosmo_a;
-      if (rank >= 2) hy *= cosmo_a;
-      if (rank >= 3) hz *= cosmo_a;
-    
-      const double dt = alpha_ * block->dt() / cosmo_a;
-
-      // Scale mass by volume if particle value is mass instead of density
-
-      int level = block->level();
-
-      // Required for Cosmology ("mass" is mass)
-      // Not for Collapse ("mass" is density)
-      if (cosmology) {
-        dens *= std::pow(2.0,rank*level);
-      }
-
-      // Accumulated single velocity array for Baryon deposit
-
+      // Loop over batches
       for (int ib=0; ib<particle.num_batches(it); ib++) {
-
+	// Pointer to array of particle masses
+	enzo_float * pmass = NULL;
         const int np = particle.num_particles(it,ib);
+
+	// Particle masses can be either a constant value for all particles,
+	// or an attribrute array
+	if (particle.has_constant (it,"mass")){
+
+	  // Check if particle also has a mass attribute
+	  ASSERT1("EnzoMethodPmDeposit::compute",
+		  "Particle type %s has both a constant and an attribute called"
+		  "'mass'. Exiting.", particle.type_name(it).c_str(),
+		  particle.has_attribute(it,"mass"));
+	  
+	  // In this case we fill an array of length np with the constant
+	  // value
+	  pmass = new enzo_float[np];
+	  im = particle.constant_index(it,"mass");
+	  for (int ip = 0; ip<np; ip++)
+	    pmass[ip] = *((enzo_float *)(particle.constant_value (it,im)));
+          
+        } else if (particle.has_attribute(it,"mass")) {
+
+	  // In this case we set to pointer to the attribute array
+	  im = particle.attribute_index(it,"mass");
+	  pmass = (enzo_float *) particle.attribute_array( it, im, ib);
+        }
+	else {
+	  ERROR1("EnzoMethodPmDeposit::compute",
+		"Particle type %s has neither a constant nor an attribute" 
+                "called 'mass', yet is in the 'has_mass' group",
+		 particle.type_name(it).c_str());
+	}
+	
+	// If mass is a constant, we simply loop through the pmass
+	// array, and so dm = 1. If its an attribute, we need to get
+	// the stride length
+	const int dm = particle.has_attribute(it,"mass") ?
+	               particle.stride(it,im) : 1;
+
+
+	// Allocate particle masses to the grid with CIC scheme
 
         if (rank == 1) {
 
@@ -180,6 +216,8 @@ void EnzoMethodPmDeposit::compute ( Block * block) throw()
 
           for (int ip=0; ip<np; ip++) {
 
+	    // Stefan: Unsure about this line, doesn't match what's in the
+	    // Enzo paper
             double x = xa[ip*dp] + vxa[ip*dv]*dt;
 
             double tx = nx*(x - xm) / (xp - xm) - 0.5;
@@ -191,8 +229,8 @@ void EnzoMethodPmDeposit::compute ( Block * block) throw()
             double x0 = 1.0 - (tx - floor(tx));
             double x1 = 1.0 - x0;
 
-            de_p[ix0] += dens*x0;
-            de_p[ix1] += dens*x1;
+            de_p[ix0] += pmass[ip*dm]/inv_vol*x0;
+            de_p[ix1] += pmass[ip*dm]/inv_vol*x1;
 
           }
 
@@ -215,7 +253,7 @@ void EnzoMethodPmDeposit::compute ( Block * block) throw()
 
           const int dp =  particle.stride(it,ia_x);
           const int dv =  particle.stride(it,ia_vx);
-
+ 
           for (int ip=0; ip<np; ip++) {
 
             double x = xa[ip*dp] + vxa[ip*dv]*dt;
@@ -236,30 +274,31 @@ void EnzoMethodPmDeposit::compute ( Block * block) throw()
             double x1 = 1.0 - x0;
             double y1 = 1.0 - y0;
 
-            if ( dens < 0.0) {
-              CkPrintf ("%s:%d ERROR: dens = %f\n", __FILE__,__LINE__,dens);
+            if ( pmass[ip*dm]/inv_vol < 0.0) {
+              CkPrintf ("%s:%d ERROR: pmass[%d]/inv_vol = %f\n",
+			__FILE__,__LINE__,ip*dm,pmass[ip*dm]/inv_vol);
             }
 
-            de_p[ix0+mx*iy0] += dens*x0*y0;
-            de_p[ix1+mx*iy0] += dens*x1*y0;
-            de_p[ix0+mx*iy1] += dens*x0*y1;
-            de_p[ix1+mx*iy1] += dens*x1*y1;
+            de_p[ix0+mx*iy0] += pmass[ip*dm]/inv_vol*x0*y0;
+            de_p[ix1+mx*iy0] += pmass[ip*dm]/inv_vol*x1*y0;
+            de_p[ix0+mx*iy1] += pmass[ip*dm]/inv_vol*x0*y1;
+            de_p[ix1+mx*iy1] += pmass[ip*dm]/inv_vol*x1*y1;
 
             if ( de_p[ix0+mx*iy0] < 0.0) {
               CkPrintf ("%s:%d ERROR: de_p %d %d = %f\n",
-                        __FILE__,__LINE__,ix0,iy0,dens);
+                        __FILE__,__LINE__,ix0,iy0,pmass[ip*dm]/inv_vol);
             }
             if ( de_p[ix1+mx*iy0] < 0.0) {
               CkPrintf ("%s:%d ERROR: de_p %d %d = %f\n",
-                        __FILE__,__LINE__,ix1,iy0,dens);
+                        __FILE__,__LINE__,ix1,iy0,pmass[ip*dm]/inv_vol);
             }
             if ( de_p[ix0+mx*iy1] < 0.0) {
               CkPrintf ("%s:%d ERROR: de_p %d %d = %f\n",
-                        __FILE__,__LINE__,ix0,iy1,dens);
+                        __FILE__,__LINE__,ix0,iy1,pmass[ip*dm]/inv_vol);
             }
             if ( de_p[ix1+mx*iy1] < 0.0) {
               CkPrintf ("%s:%d ERROR: de_p %d %d = %f\n",
-                        __FILE__,__LINE__,ix1,iy1,dens);
+                        __FILE__,__LINE__,ix1,iy1,pmass[ip*dm]/inv_vol);
             }
           }
 
@@ -318,19 +357,23 @@ void EnzoMethodPmDeposit::compute ( Block * block) throw()
             double y1 = 1.0 - y0;
             double z1 = 1.0 - z0;
 
-            de_p[ix0+mx*(iy0+my*iz0)] += dens*x0*y0*z0;
-            de_p[ix1+mx*(iy0+my*iz0)] += dens*x1*y0*z0;
-            de_p[ix0+mx*(iy1+my*iz0)] += dens*x0*y1*z0;
-            de_p[ix1+mx*(iy1+my*iz0)] += dens*x1*y1*z0;
-            de_p[ix0+mx*(iy0+my*iz1)] += dens*x0*y0*z1;
-            de_p[ix1+mx*(iy0+my*iz1)] += dens*x1*y0*z1;
-            de_p[ix0+mx*(iy1+my*iz1)] += dens*x0*y1*z1;
-            de_p[ix1+mx*(iy1+my*iz1)] += dens*x1*y1*z1;
+            de_p[ix0+mx*(iy0+my*iz0)] += pmass[ip*dm]/inv_vol*x0*y0*z0;
+            de_p[ix1+mx*(iy0+my*iz0)] += pmass[ip*dm]/inv_vol*x1*y0*z0;
+            de_p[ix0+mx*(iy1+my*iz0)] += pmass[ip*dm]/inv_vol*x0*y1*z0;
+            de_p[ix1+mx*(iy1+my*iz0)] += pmass[ip*dm]/inv_vol*x1*y1*z0;
+            de_p[ix0+mx*(iy0+my*iz1)] += pmass[ip*dm]/inv_vol*x0*y0*z1;
+            de_p[ix1+mx*(iy0+my*iz1)] += pmass[ip*dm]/inv_vol*x1*y0*z1;
+            de_p[ix0+mx*(iy1+my*iz1)] += pmass[ip*dm]/inv_vol*x0*y1*z1;
+            de_p[ix1+mx*(iy1+my*iz1)] += pmass[ip*dm]/inv_vol*x1*y1*z1;
 
-          }
-        }
-      }
-    }
+          } // Loop over particle in batch
+        } // if rank == 3
+	
+	// If constant mass, delete the pmass array
+	if (particle.has_constant(it,"mass")) delete [] pmass;
+
+      } // Loop over batches
+    } // Loop over particle types in 'has_mass' group
 
     //--------------------------------------------------
     // Add gas density
--- a/src/Enzo/enzo_EnzoMethodPmDeposit.cpp
+++ b/src/Enzo/enzo_EnzoMethodPmDeposit.cpp
@@ -2,7 +2,7 @@
 
 /// @file     enzo_EnzoMethodPmDeposit.cpp
 /// @author   James Bordner (jobordner@ucsd.edu)
-/// @author   Stefan Arridge (stefan.arridge@gmail.com)
+/// Edited:   Stefan Arridge (stefan.arridge@gmail.com)
 /// @date     Fri Apr  2 17:05:23 PDT 2010
 /// @brief    Implements the EnzoMethodPmDeposit class
 ///
@@ -117,19 +117,11 @@ void EnzoMethodPmDeposit::compute ( Block * block) throw()
 					  block->time() + alpha_*block->dt());
     }
 
-    // Stefan: Not sure if cell widths need to be converted to a physical
-    // length
-    //if (rank >= 1) hx *= cosmo_a;
-    //if (rank >= 2) hy *= cosmo_a;
-    //if (rank >= 3) hz *= cosmo_a;
-
     double inv_vol = 1.0;
     if (rank >= 1) inv_vol /= hx;
     if (rank >= 2) inv_vol /= hy;
     if (rank >= 3) inv_vol /= hz;
 
-    // Stefan: Unsure of the meaning of this line. How is cosmological time
-    // defined in Enzo?
     const double dt = alpha_ * block->dt() / cosmo_a;
 
     // Get the number of particle types in the 'has_mass' group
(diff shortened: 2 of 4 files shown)


src/Enzo/enzo_EnzoMethodPmDeposit.cpp
```diff
@@ -2,6 +2,7 @@
 
 /// @file     enzo_EnzoMethodPmDeposit.cpp
 /// @author   James Bordner (jobordner@ucsd.edu)
+/// Edited:   Stefan Arridge (stefan.arridge@gmail.com)
```
mabruzzo, line 5:
```suggestion
/// @author:   Stefan Arridge (stefan.arridge@gmail.com)
```
I would encourage you to use `@author` here since that's a keyword recognized by Doxygen.

src/Enzo/enzo_EnzoMethodPmDeposit.cpp
```diff
@@ -180,6 +208,9 @@ void EnzoMethodPmDeposit::compute ( Block * block) throw()
 
           for (int ip=0; ip<np; ip++) {
 
+	    // Stefan: Unsure about this line, doesn't match what's in the
```
DeovratPrasad, line 211: This needs to be looked at ASAP by someone who is the expert at evolving 'Particles'

src/Enzo/enzo_EnzoMethodPmDeposit.cpp
```diff
@@ -180,6 +208,9 @@ void EnzoMethodPmDeposit::compute ( Block * block) throw()
...
+	    // Enzo paper, which says that particles are drifted by half a timestep
+	    // before being deposited to the grid
             double x = xa[ip*dp] + vxa[ip*dv]*dt;
```
DeovratPrasad, line 214: In the light of the comment above by Stefan, this needs to be looked at ASAP before approving, preferrably by someone who is the expert at evolving 'Particles'.

DeovratPrasad, line 214: In the light of the comment above by Stefan, this needs to be looked at ASAP before approving, preferrably by someone who is the expert at evolving 'Particles'.

DeovratPrasad (commented):
someone who deals with particles algo needs to look at the comment by Stefan in file enzo_EnzoMethodPmDeposit.cpp . I have also left a comment there flagging the issue.

stefanarridge:
> someone who deals with particles algo needs to look at the comment by Stefan in file enzo_EnzoMethodPmDeposit.cpp . I have also left a comment there flagging the issue.

Thanks for pointing that out. That comment was due to a misunderstanding on my part. I'll delete it now.

src/Cello/data_ParticleDescr.cpp
```diff
@@ -253,6 +253,24 @@ int ParticleDescr::attribute_index (int it, std::string attribute_name) const
 
 //----------------------------------------------------------------------
 
+bool ParticleDescr::has_attribute(int it, std::string attribute_name) const
+{
+  ASSERT1("ParticleDescr::is_attribute",
```
jwise77, line 258: is -> has

src/Cello/data_ParticleDescr.cpp
```diff
@@ -253,6 +253,24 @@ int ParticleDescr::attribute_index (int it, std::string attribute_name) const
 
 //----------------------------------------------------------------------
 
+bool ParticleDescr::has_attribute(int it, std::string attribute_name) const
```
jwise77, line 256: This was included in #49 as `is_attribute`

jwise77 (commented):
Nearly all of these changes look good to me. See my two comments. #49 included the `has_attribute` method as `is_attribute`. I can approve once these instances are changed to `is_attribute` (probably easier to make changes in this PR rather than the #49 changes) and the conflicts are fixed. Thanks!

bwoshea:
@stefanarridge it looks like this PR has some conflicts, as well as some comments.  Would mind fixing the conflicts, addressing the comments, and then letting the reviewers know it's ready for review?  Thank you!

stefanarridge:
> Nearly all of these changes look good to me. See my two comments. #49 included the has_attribute method as is_attribute. I can approve once these instances are changed to is_attribute (probably easier to make changes in this PR rather than the #49 changes) and the conflicts are fixed. Thanks!

I names has_attribute as such to maintain consistency with the has_constant method. I have now decided to name the methods 'is_attribute' and 'is_constant' instead.

src/Cello/data_ParticleDescr.cpp
stefanarridge: I named has_attribute as such to maintain consistency with the has_constant method. I have now decided to name the methods 'is_attribute' and 'is_constant' instead.

src/Enzo/enzo_EnzoMethodPmDeposit.cpp
```diff
@@ -252,53 +256,49 @@ void EnzoMethodPmDeposit::compute ( Block * block) throw()
...
+	    
+      	    de_p[ix0+mx*iy0] += pmass[ip*dm]*inv_vol*x0*y0;
+      	    de_p[ix1+mx*iy0] += pmass[ip*dm]*inv_vol*x1*y0;
+      	    de_p[ix0+mx*iy1] += pmass[ip*dm]*inv_vol*x0*y1;
+      	    de_p[ix1+mx*iy1] += pmass[ip*dm]*inv_vol*x1*y1;
+	    
       	    if ( de_p[ix0+mx*iy0] < 0.0) {
       	      CkPrintf ("%s:%d ERROR: de_p %d %d = %f\n",
-      	   	        __FILE__,__LINE__,ix0,iy0,dens);
+      	   	        __FILE__,__LINE__,ix0,iy0,inv_vol);
```
jwise77, line 287: Instead of `inv_vol` that's always positive, it might be more informative to report on the negative density (`de_p[ix0+mx*iy0]`) and/or mass (`pmass[ip*dm]`). Applies to all instances below, too.

From the previous version, I also see that this error check only happens in 2D. It should be consistently checked (or not) regardless of the rank. It doesn't need to be fixed in this PR, but I'll make an issue.

jwise77 (approved):
All looks good! I only had a minor comment that isn't related to the density -> mass changes.

src/Enzo/enzo_EnzoMethodPmDeposit.cpp
```diff
@@ -117,45 +118,41 @@ void EnzoMethodPmDeposit::compute ( Block * block) throw()
...
+					  block->time() + alpha_*block->dt());
     }
-    if (rank >= 1) hx *= cosmo_a;
-    if (rank >= 2) hy *= cosmo_a;
-    if (rank >= 3) hz *= cosmo_a;
+
+    double inv_vol = 1.0;
+    if (rank >= 1) inv_vol /= hx;
+    if (rank >= 2) inv_vol /= hy;
+    if (rank >= 3) inv_vol /= hz;
```
mabruzzo, line 135: If my previous comment (about multiplying the cell widths by the scale factor) is wrong, then I have a minor stylistic suggestion. 

It might make the code easier to follow if you moved this block of code up to line 121 (before you handle the cosmology stuff). There are 2 benefit to this:
1. The only usage of `hx`, `hy`, and `hz` would occur right after the values are retrieved.
2. The only usage of the cosmology information is used (i.e. the line `const double dt = alpha_ * block->dt() / cosmo_a;`) would now occur right after the cosmology information is retrieved.

Of course, if you don't think that this improves things, feel free to ignore this

src/Enzo/enzo_EnzoMethodPmDeposit.cpp
```diff
@@ -166,84 +163,91 @@ void EnzoMethodPmDeposit::compute ( Block * block) throw()
...
+		  "Particle type %s has both a constant and an attribute called"
+		  "'mass'. Exiting.", particle.type_name(it).c_str(),
+		  particle.is_attribute(it,"mass"));
+	  
+	  // In this case we fill an array of length np with the constant
+	  // value
+	  pmass = new enzo_float[np];
+	  im = particle.constant_index(it,"mass");
+	  for (int ip = 0; ip<np; ip++)
+	    pmass[ip] = *((enzo_float *)(particle.constant_value (it,im)));
```
mabruzzo, line 187: It's a little unfortunate that we allocate, fill, and free the `pmass` buffer for every batch of each particle type with a constant mass. It would probably be better if we did just did this once for each particle type (with a constant mass) using the max batch size of that particle.

I realize that you just copied the existing style, so feel free to ignore this recommendation.

src/Enzo/enzo_EnzoMethodPmDeposit.cpp
```diff
@@ -117,45 +118,41 @@ void EnzoMethodPmDeposit::compute ( Block * block) throw()
...
-      double time = block->time();
-      double dt   = block->dt();
-      cosmology->compute_expansion_factor (&cosmo_a,&cosmo_dadt,time+alpha_*dt);
-
+      cosmology->compute_expansion_factor(&cosmo_a,&cosmo_dadt,
+					  block->time() + alpha_*block->dt());
     }
-    if (rank >= 1) hx *= cosmo_a;
-    if (rank >= 2) hy *= cosmo_a;
-    if (rank >= 3) hz *= cosmo_a;
```
mabruzzo, line 143: It seems like that removing this is going to affect the way that the gas density gets deposited (which happens at the end of this function).

If this is true, then it's a little disturbing that none of our automated tests seem to detect this.

mabruzzo (changes requested):
This looks great! I've got a couple of minor suggestions. I'm a little concerned that one of your changes might have broken gas density deposition in cosmological simulations (there is a comment highlighting that in `enzo_EnzoMethodPmDeposit`). But I could definitely be wrong.

Apologies for not providing these suggestions sooner. I wasn't originally intending to fully review this because I didn't really understand the Particle machinery (but James's presentation at the workshop made things a lot clearer).

src/Enzo/enzo_EnzoMethodPmDeposit.cpp
stefanarridge: Thanks for pointing this out, I think this is an artifact of some find-replaces I did.

src/Cello/data_ParticleDescr.cpp
```diff
@@ -256,10 +256,10 @@ int ParticleDescr::attribute_index (int it, std::string attribute_name) const
 bool ParticleDescr::is_attribute(int it, std::string attribute_name) const
 {
   ASSERT1("ParticleDescr::is_attribute",
-          "Trying to access unknown particle type %d",
-          it,
-          (0 <= it && it < num_types()));
-
+          "Trying to access unknown particle type '%s'",
+          type_name(it).c_str(),
```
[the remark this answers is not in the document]
stefanarridge: I think you're right. I'll change it.

src/Enzo/enzo_EnzoMethodPmDeposit.cpp
stefanarridge: Okay, I'll look into this. I was under the impression that that both gas density and particle density were comoving quantities. I suppose if a physical length scale is required for depositing the gas density, then I can define two separate variables i.e hx_com and hy_phys, or something similar.

src/Enzo/enzo_EnzoMethodPmDeposit.cpp
stefanarridge: Good suggestion, I'll look into it.

src/Enzo/enzo_EnzoMethodPmDeposit.cpp
mabruzzo: I was just trying to point out that this change was going to modify the deposititon of the gas density.

You could definitely be correct (I know basically nothing about the units of cosmological simulations). In that case the old implementation must have been wrong.

Could somebody familiar with the density units used for cosmological simulations in Enzo quickly clarify this for us? @jwise77 @bwoshea  @gregbryan

stefanarridge: Yes, I know what you mean. It would be good to understand which quantities need to be passed to dep_grid_cic (i.e., physical or comoving). I am also concerned that the value of dtf which is passed is not correct. Currently it is set to alpha_, which is a dimensionless parameter which is 0.5 by default, rather than a quantity with dimensions of time.

gregbryan: In Enzo, the particle "mass" is actually a comoving density, adjusted for the level that it is on, so that when it gets deposited to the GravitatingMassField (which is the comoving density field used to calculate the gravitational acceleration), there is no multiplication by a volume. Here, I'm not quite sure what should be done -- the mass of a particle is a constant, so if you wanted to compute a comoving density, then you would want to divide by the comoving volume of the cell.  Normally dx is comoving, so a factor of the scale factor is not required, but I have to admit I have not really looked at what is going on in this routine, so best to take what I say with a grain of salt.

stefanarridge: @gregbryan , the purpose of this PR is to redefine the "mass" attribute, so that it is always treated as mass, rather than (sometimes) as density. As far as I know, in a cosmological simulation, length scales and densities are all comoving quantities, which is why I deleted the lines which multiply hx,hy,hz by "cosmo_a", so that, when the mass is multiplied by 1/(hx * hy * hz), the output is a comoving density.

The issue which @mabruzzo pointed out, is that this will affect the deposition of gas density, which is done by calling the 'dep_grid_cic' function:

`
FORTRAN_NAME(dep_grid_cic)(de,de_gas,temp,
  
  
      			       vx, vy, vz,
  
  
      			       &amp;dtf, rfield, &amp;rank,
  
  
      			       &amp;hxf,&amp;hyf,&amp;hzf,
  
  
      			       &amp;mx,&amp;my,&amp;mz,
  
  
      			       &amp;gxi,&amp;gyi,&amp;gzi,
  
  
      			       &amp;nxi,&amp;nyi,&amp;nzi,
  
  
      			       &amp;i0,&amp;i0,&amp;i0,
  
  
      			       &amp;nx,&amp;ny,&amp;nz,
  
  
      			       &amp;i1,&amp;i1,&amp;i1);`
      			       
Should hxf, hyf, hzf be comoving or physical length scales? 

I'm also surprised that dtf is set equal to 'alpha_'. Should it not be the same as the 'dt' used to drift the particles, i.e.:
`const double dt = alpha_ * block->dt() / cosmo_a;`

mabruzzo: I took a look at ``dep_grid_cic``, and I think that it's effectively just advancing the density over `dt` (i.e. the analogous operation to drifting the particles). I'm also pretty sure that the output density `de_gas` has the same units as the input `de`
- As an aside, I'm pretty sure that a lot of these arguments that get passed have to do with complications that arise from the patch-based AMR used in the original _Enzo_.

If I'm remembering correctly, the fields used to store gas properties store values in comoving units for cosmological simulations, so `hxf`, `hyf`, and `hzf` should probably be comoving.

Furthermore, since the `"density_total"` field is the sum of the `"density"` field (after drifting) and the `"density_particle"` field, I suspect that `"density_total"`, `"density_particle"` and `"density_particle_accumulate"` should all be comoving densities (in cosmological simulations).

I could be wrong, but this makes the most sense to me. The easiest way to check this would be to write some tests.

src/Enzo/enzo_EnzoMethodPmDeposit.cpp
```diff
@@ -162,85 +198,109 @@ void EnzoMethodPmDeposit::compute ( Block * block) throw()
...
+	    
       	    double x = xa[ip*dp] + vxa[ip*dv]*dt;
 
       	    double tx = nx*(x - xm) / (xp - xm) - 0.5;
-
-        	  int ix0 = gx + floor(tx);
-
+	    
+	    int ix0 = gx + floor(tx);
+
```
mabruzzo, line 276:
```suggestion

	    int ix0 = gx + floor(tx);

```

mabruzzo: There's also something funky about the indentation

src/Enzo/enzo_EnzoMethodPmDeposit.cpp
```diff
@@ -162,85 +198,109 @@ void EnzoMethodPmDeposit::compute ( Block * block) throw()
...
       	    double x1 = 1.0 - x0;
+	    
+            de_p[ix0] += pdens[ip*stride]*x0;
+            de_p[ix1] += pdens[ip*stride]*x1;
 
-      	    de_p[ix0] += pdens[ip]*dens*x0;
-      	    de_p[ix1] += pdens[ip]*dens*x1;
-
-      	  } // np
+	    if (de_p[ix0] < 0.0) {
```
mabruzzo, line 285:
```suggestion
      	    if (de_p[ix0] < 0.0) {
```

src/Enzo/enzo_EnzoMethodPmDeposit.cpp
```diff
@@ -162,85 +198,109 @@ void EnzoMethodPmDeposit::compute ( Block * block) throw()
...
+      	      CkPrintf ("%s:%d ERROR: de_p %d = %f\n",
+      	   	        __FILE__,__LINE__,ix0,de_p[ix0]);
+      	    }
 
+	    if (de_p[ix1] < 0.0) {
```
mabruzzo, line 290:
```suggestion
      	    if (de_p[ix1] < 0.0) {
```

src/Enzo/enzo_EnzoMethodPmDeposit.cpp
```diff
@@ -162,85 +198,109 @@ void EnzoMethodPmDeposit::compute ( Block * block) throw()
...
+	    
+            de_p[ix0] += pdens[ip*stride]*x0;
+            de_p[ix1] += pdens[ip*stride]*x1;
 
-      	    de_p[ix0] += pdens[ip]*dens*x0;
-      	    de_p[ix1] += pdens[ip]*dens*x1;
-
-      	  } // np
+	    if (de_p[ix0] < 0.0) {
+      	      CkPrintf ("%s:%d ERROR: de_p %d = %f\n",
```
mabruzzo, line 286: Totally up to you, but it might make sense to  use the `WARNING2` macro here.

src/Enzo/enzo_EnzoMethodPmDeposit.cpp
```diff
@@ -162,85 +198,109 @@ void EnzoMethodPmDeposit::compute ( Block * block) throw()
...
+      	  } // Loop over particles in batch
+	  
         } else if (rank == 2) {
 
       	  const int ia_x  = particle.attribute_index(it,"x");
       	  const int ia_y  = particle.attribute_index(it,"y");
       	  const int ia_vx = particle.attribute_index(it,"vx");
       	  const int ia_vy = particle.attribute_index(it,"vy");
-
+
```
mabruzzo, line 303:
```suggestion

```

src/Enzo/enzo_EnzoMethodPmDeposit.cpp
```diff
@@ -249,53 +309,49 @@ void EnzoMethodPmDeposit::compute ( Block * block) throw()
 
       	  const int dp =  particle.stride(it,ia_x);
       	  const int dv =  particle.stride(it,ia_vx);
-
-	        for (int ip=0; ip<np; ip++) {
-
+	  
+	  for (int ip=0; ip<np; ip++) {
+
```
mabruzzo, line 314:
```suggestion

	  for (int ip=0; ip<np; ip++) {

```

src/Enzo/enzo_EnzoMethodPmDeposit.cpp
```diff
@@ -249,53 +309,49 @@ void EnzoMethodPmDeposit::compute ( Block * block) throw()
...
+	  
+	  for (int ip=0; ip<np; ip++) {
+	    
       	    double x = xa[ip*dp] + vxa[ip*dv]*dt;
       	    double y = ya[ip*dp] + vya[ip*dv]*dt;
 
       	    double tx = nx*(x - xm) / (xp - xm) - 0.5;
       	    double ty = ny*(y - ym) / (yp - ym) - 0.5;
-
+
```
mabruzzo, line 320:
```suggestion

```

src/Enzo/enzo_EnzoMethodPmDeposit.cpp
```diff
@@ -249,53 +309,49 @@ void EnzoMethodPmDeposit::compute ( Block * block) throw()
...
       	    double y = ya[ip*dp] + vya[ip*dv]*dt;
 
       	    double tx = nx*(x - xm) / (xp - xm) - 0.5;
       	    double ty = ny*(y - ym) / (yp - ym) - 0.5;
-
+	    
       	    int ix0 = gx + floor(tx);
       	    int iy0 = gy + floor(ty);
-
+
```
mabruzzo, line 323:
```suggestion

```

src/Enzo/enzo_EnzoMethodPmDeposit.cpp
```diff
@@ -249,53 +309,49 @@ void EnzoMethodPmDeposit::compute ( Block * block) throw()
...
-
+	    
       	    int ix0 = gx + floor(tx);
       	    int iy0 = gy + floor(ty);
-
+	    
       	    int ix1 = ix0 + 1;
       	    int iy1 = iy0 + 1;
-
+
```
mabruzzo, line 326:
```suggestion

```

src/Enzo/enzo_EnzoMethodPmDeposit.cpp
```diff
@@ -249,53 +309,49 @@ void EnzoMethodPmDeposit::compute ( Block * block) throw()
...
-
+	    
       	    int ix1 = ix0 + 1;
       	    int iy1 = iy0 + 1;
-
+	    
       	    double x0 = 1.0 - (tx - floor(tx));
       	    double y0 = 1.0 - (ty - floor(ty));
-
+
```
mabruzzo, line 329:
```suggestion

```

src/Enzo/enzo_EnzoMethodPmDeposit.cpp
```diff
@@ -305,7 +361,7 @@ void EnzoMethodPmDeposit::compute ( Block * block) throw()
       	  const int ia_vx = particle.attribute_index(it,"vx");
       	  const int ia_vy = particle.attribute_index(it,"vy");
       	  const int ia_vz = particle.attribute_index(it,"vz");
-
+
```
mabruzzo, line 364:
```suggestion

```

src/Enzo/enzo_EnzoMethodPmDeposit.cpp
mabruzzo: Without doing a deep-dive back into this, it's unclear to me if you've made any changes on this front. If you have made changes, I'll take another look

src/Enzo/enzo_EnzoMethodPmDeposit.cpp
```diff
@@ -115,45 +138,58 @@ void EnzoMethodPmDeposit::compute ( Block * block) throw()
...
+    double density_scale_factor = 1.0;
+    for (int j = 0; j < rank*level; j++)
+      density_scale_factor *= 2.0;
+    
+    // Get cosmological scale factors, if cosmology is turned on
     enzo_float cosmo_a=1.0;
     enzo_float cosmo_dadt=0.0;
     EnzoPhysicsCosmology * cosmology = enzo::cosmology();
-
+
```
mabruzzo, line 159:
```suggestion

```

mabruzzo (commented):
The vast majority of my recommended edits relate to trailing whitespace.

It's not totally obvious to me whether you made more changes related to depositing the gas density. Let me know if you did and I'll take another look at that.

stefanarridge:
> The vast majority of my recommended edits relate to trailing whitespace.

I know this is a bit of a recurring problem with my PRs, as well as indentation issues. Is there a particular linter that you would recommend @mabruzzo?

peeples:
Hey @mabruzzo can you take look at this / are there other changes needed before this is merged?

mabruzzo:
> Hey @mabruzzo can you take look at this / are there other changes needed before this is merged?

I'll look this afternoon

doc/source/user/problem_method.rst
```diff
@@ -518,12 +518,27 @@ fields
 
 particles
 ---------
+
+For a given particle type to be deposited to the total density field,
+it must be part of the ``"is_gravitating"`` group, and must have
+exactly one of the following: an attribute called ``"mass"``, a
+constant called ``"mass"``, an attribute called ``"density"``, or
+a constant called ``"density"``.
+
```
fearmayo, line 527: The next two paragraphs appear at odds with one another. For the 'mass' case the density of the particle in the cell is found. However for the 'density' case it is something different. In the 'density' case it looks like the density is being increased to account for cells which are refined (in a unigrid run nothing changes). It seems to me that this will only work if the 'density' is the mass divided by the volume of the root grid cell? Is that correct?

mabruzzo (approved):
Sorry for taking so long to come back to this.

Looks good to me!

stefanarridge:
Even though this now has two approvals, I suggest holding off on merging this just yet, at least until @brittonsmith checks whether this breaks anything in yt. I have also also asked Will to merge this into his STARSS branch and check whether everything runs smoothly.

jwise77:
In the meantime, I'll give your changes another review to give feedback on your updates.

jwise77 (approved):
All looks good, but I will comment on the `dtf = _alpha` factor in a reply to John R's comment.

src/Enzo/enzo_EnzoMethodPmDeposit.cpp
jwise77: > I'm also surprised that dtf is set equal to 'alpha_'. Should it not be the same as the 'dt' used to drift the particles, i.e.:
const double dt = alpha_ * block->dt() / cosmo_a;

I agree that it should be the same dt as the particle drift, but in Enzo, dt is [actually zero](https://github.com/enzo-project/enzo-dev/blob/main/src/enzo/Grid_DepositBaryons.C#L198) for the PPM solver but not the RK2 solver ([diff from 2011](https://github.com/enzo-project/enzo-dev/commit/c16a9eb743d4716f487ac0ace62e4367d65b54aa#diff-0264e55b281d53b17e54c49abf1658d7ce6a19db6fe03686b7d225f1e9049908L174)).

I looked at a very old version of Enzo from 2005, and I found this comment that using a non-zero dt creates an instability. I think we should change it to zero to be consistent with Enzo.

```
    float dt = (DepositTime - Time)/a;
    dt = 0; /* the previous line is correct but creates an instability */
```

brittonsmith:
Sorry for the late comments here. Firstly, I don't foresee any issues with the yt frontend. The existence or absence of the "is_gravitating" attribute makes for an easy check for whether the data was made before and after this change.

I apologize if this has already been discussed and if so I'm happy to leave it. Supporting both density and mass particles seems unnecessary. Is there any benefit to defining particle masses as densities beyond skipping a multiplication/division? It seems like we are making things more complicated for ourselves by continuing to support both options.

stefanarridge:
> I apologize if this has already been discussed and if so I'm happy to leave it. Supporting both density and mass particles seems unnecessary. Is there any benefit to defining particle masses as densities beyond skipping a multiplication/division? It seems like we are making things more complicated for ourselves by continuing to support both options.

There is in face little to no computational benefit in using densities rather than masses: in the "density" case, the densities have to be multiplied by 2^(rank * level) (rank is the number of dimensions, level is the refinement level), while in the "mass" case, the masses have to be multiplied by the inverse cell volume.

 In principle, I agree with you, it would be better not to have the "density" option, but I am allowing for this option in order to make it easier for users to convert their old input parameter files or IC files. For example, if you look at the changes made to `input/Cosmology/method_cosmology,incl`, all that was required was to change the name of the `"mass"` constant to `"density"` (and change the name of the `"has_mass"` group to `"is_gravitating"`). The alternative would have been to change the value of the `"mass"` constant by multiplying it by the root level cell volume.

brittonsmith:
> In principle, I agree with you, it would be better not to have the "density" option, but I am allowing for this option in order to make it easier for users to convert their old input parameter files or IC files. For example, if you look at the changes made to `input/Cosmology/method_cosmology,incl`, all that was required was to change the name of the `"mass"` constant to `"density"` (and change the name of the `"has_mass"` group to `"is_gravitating"`). The alternative would have been to change the value of the `"mass"` constant by multiplying it by the root level cell volume.

I think this is an opportunity for us to rip off a relatively painless bandaid. Since the density option still requires a renormalization to the local grid cell size, this doesn't even work the way it does in enzo-dev, so this could be a potential point of confusion for analysis or if more code is ported from that codebase. My personal opinion is that leaving the density option in leaves the door open for bugs, but I would like to hear other opinions. @bwoshea, @jwise77, @gregbryan, @fearmayo, @peeples, @mabruzzo  ?

src/Enzo/enzo_EnzoMethodPmDeposit.cpp
gregbryan: OK, I looked this over and I *think* removing the multiplication by cosmo_a is the right thing to do. As others have pointed out, hx,hy,hz should be comoving, so this is not necessary.  Regarding alpha_, this is for the particles, and I think that is only set to zero for the baryons, so OK to keep as is?

src/Enzo/enzo_EnzoMethodPmDeposit.cpp
```diff
@@ -391,9 +490,16 @@ void EnzoMethodPmDeposit::compute ( Block * block) throw()
...
+    /* Stefan: Not sure if these next four lines are correct.
+       I include the factors of cosmo_a for consistency with 
+       previous version. Also, not sure why dtf is
+       initialised with the value of alpha_.
+     */
+    
+    enzo_float hxf = hx * cosmo_a;
+    enzo_float hyf = hy * cosmo_a;
+    enzo_float hzf = hz * cosmo_a;
```
gregbryan, line 502: I think this cosmo_a factors should not be here, for the same reason discussed earlier.

gregbryan:
> > In principle, I agree with you, it would be better not to have the "density" option, but I am allowing for this option in order to make it easier for users to convert their old input parameter files or IC files. For example, if you look at the changes made to `input/Cosmology/method_cosmology,incl`, all that was required was to change the name of the `"mass"` constant to `"density"` (and change the name of the `"has_mass"` group to `"is_gravitating"`). The alternative would have been to change the value of the `"mass"` constant by multiplying it by the root level cell volume.
> 
> I think this is an opportunity for us to rip off a relatively painless bandaid. Since the density option still requires a renormalization to the local grid cell size, this doesn't even work the way it does in enzo-dev, so this could be a potential point of confusion for analysis or if more code is ported from that codebase. My personal opinion is that leaving the density option in leaves the door open for bugs, but I would like to hear other opinions. @bwoshea, @jwise77, @gregbryan, @fearmayo, @peeples, @mabruzzo ?

I also think it may be simpler just to only allow mass. I always regretted the Enzo decision to make it density (the level of confusion added was not worth the small gains). I do know that you have already invested quite a bit of effort, so would completely understand punting on that for now.

I think the bigger issue is that we need a test problem to verify the choices made. For example, there is a discussion about using hx or hx * cosmo_a. I think we have arrived at the right thing, but there is only one answer that is right and it would be great to be able to show it. For cosmology, one could test the growth rate of the Zeldovich pancake or the power spectrum for a cosmo run. For non-cosmology, maybe just the growth timescale for the collapse problem?

stefanarridge:
> I also think it may be simpler just to only allow mass. I always regretted the Enzo decision to make it density (the level of confusion added was not worth the small gains). I do know that you have already invested quite a bit of effort, so would completely understand punting on that for now.

It won't be much work to get rid of the "density" option. However, users will have to update parameter files / IC files which set a value for particle "mass", although it should be as simple as just multiplying by the root level cell volume.

> I think the bigger issue is that we need a test problem to verify the choices made. For example, there is a discussion about using hx or hx * cosmo_a. I think we have arrived at the right thing, but there is only one answer that is right and it would be great to be able to show it. For cosmology, one could test the growth rate of the Zeldovich pancake or the power spectrum for a cosmo run. For non-cosmology, maybe just the growth timescale for the collapse problem?

I have been considering making a separate PR which will address the "gas density deposition" issue, as well as a few other issues with the gravity method which I think need to be addressed. The issue I have is that I am not completely certain what the `dep_grid_cic` function is doing. Am I right in thinking that, in the special case where the "source" and "destination" fields have the same dimensions, are on the same refinement level, and are not offset, which is the case here, that it essentially does the following:

- Loop over each active cell in the source field (in this case `"density"`)
- Take the position corresponding to the center of the cell, and compute a new position which is this position plus "velocity times dt", where "velocity" is the value of the velocity field in this cell, and "dt" is the suitable timestep.
- Use this new position to do CIC deposition onto the destination field (a new field called `de_gas` in this case, but may as well be the `"density_total"` field).

The way that I think about it is that it is essentially depositing mass in the same way as is done for the particles, only this time, the "particle positions" are the cell centers, the "particle velocities" are the values of the velocity field, and the "particle masses" are the cell masses (density times cell volume). If this is indeed how its done, it would be relatively straightforward to write this down  in C++ code in `src/Enzo/enzo_EnzoMethodPmDeposit.cpp", which would simplify things a lot.

Obviously, how to test whether its "correct" is a separate question which I'm not sure how to answer.

stefanarridge:
Update: I have now made the following changes:

- I have removed the "density" option for gravitating particles. That is, if a particle type is in the "is_gravitating" group, then it must have either a constant or an attribute called "mass".
- I have updated `input/Cosmology/method_cosmology.incl` so that `"dark"` particles have a constant called `"mass"` rather than `"density"`, but the value of this constant has changed to reflect that it is a quantity with dimensions of mass rather than dimensions of density. This value was computed by multiplying the value of the `"density"` constant by the root level cell volume. For a cosmological simulation, where the units are defined such that the volume of the domain is 1, this is equivalent to dividing by the number of root level cells (32^3 in this case), which are specified by the `Mesh:root_size` parameter.
- the ".parameters" and ".libconfig" output files always contain a parameter in the Particle group called "mass_is_mass". This is intended to be used by the yt Enzo-E frontend, to indicate whether the data was produced by a version of Enzo-E which includes the changes introduced by this PR. I had to hack the code a bit to add in this parameter, and it doesn't work quite as intended, but there will always be a parameter called "mass_is_mass", which will almost certainly be a string with the value "dummy". The chances it isn't are 1 in 2^29... there is also a 1 in 2^31 chance that something weird might happen, and it might produce something nonsensical which yt won't be able to read in. If you're wondering what's going on, its because there is a int variable which doesn't get initialised properly, and so essentially it has a random value. If this value is outside the range 0-7, "mass_is_mass" will be "dummy"... if it's inside that range, something else will happen.

Again, I suggest this doesn't get merged in until the yt Enzo-E frontend is updated, which @brittonsmith is working on. I welcome comments and questions in the meantime.

PR #96, opened 2021-05-11
Can supply Grackle path via environment variable

clairekope:
Made a small tweak to config files that auto-searched for the Grackle install to let it be a little more flexible. The user can instead define the environment variable GRACKLE_PATH to override the results of the search. Only tested `linux_gnu`.


clairekope:
Oh! I did not know that. I didn't see any mention of GRACKLE_HOME in the docs and I didn't check the function, but defining GRACKLE_HOME does work as expected. So, I guess this PR can be deleted and and change to the docs should be made instead :)

Issue #98, opened 2021-05-19
`NumberOfBaryonFields` ptr versus int logic


pgrete:
After a system update (which came with an updated `g++`), Enzo-E now fails to compile with
```
build-master/Enzo/enzo_EnzoMethodPpml.cpp: In member function ‘virtual double EnzoMethodPpml::timestep(Block*) const’:
build-master/Enzo/enzo_EnzoMethodPpml.cpp:96:39: error: ordered comparison of pointer with integer zero (‘int*’ and ‘int’)
   96 |   if (EnzoBlock::NumberOfBaryonFields > 0) {
      |       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~
Fatal Error by charmc in directory /home/pgrete/src/enzo-e
   Command g++ -DCMK_GFORTRAN -I/home/pgrete/src/charm/bin/../include -D__CHARMC__=1 -DCONFIG_PRECISION_DOUBLE -DSMALL_INTS -DCONFIG_NODE_SIZE=64 -DCONFIG_NODE_SIZE_3=192 -DNO_FREETYPE -DCONFIG_USE_PERFORMANCE -DCONFIG_USE_MEMORY -DCONFIG_HAVE_VERSION_CONTROL -Iinclude -I/usr/include -I/usr/include/boost -I/lib/x86_64-linux-gnu/include -O3 -g -ffast-math -funroll-loops -fPIC -pedantic -U_FORTIFY_SOURCE -fno-stack-protector -fno-lifetime-dse -c build-master/Enzo/enzo_EnzoMethodPpml.cpp -o build-master/Enzo/enzo_EnzoMethodPpml.o returned error code 1
```

This also applies to other files, e.g.,  `build-master/Enzo/enzo_SetMinimumSupport.cpp` or `build-master/Enzo/enzo_SolveMHDEquations.cpp`.

As far as I can tell `int EnzoBlock::NumberOfBaryonFields[CONFIG_NODE_SIZE];` from `enzo_EnzoBlock` will always be `!= nullptr` because `CONFIG_NODE_SIZE` is a compile time constant.
Thus, the conditional as it stands will always be true (if it compiles).

I assume that the idea behind the conditional is that those code pieces should only be called if there are actualy baryon fields.
If that's true, then the check should be updated to sth like `EnzoBlock::NumberOfBaryonFields[my_processing_element] > 0`.

[...]

PR #102, opened 2021-05-19
Shorten testing

jobordner:
CircleCI tests have been failing due to timing out on individual tests.  While testing needs some more serious updating to bring everything back online again (unit tests, cosmology, etc.), this is intended as a quick fix to shorten the time of longer tests so that CircleCI doesn't time out.  With the current tests being run, that's just "adapt-L5", which does simple hydro on a 2D AMR mesh, and "mesh-balanced", which is obsolete and redundant with adapt-L5 (it doesn't test load balancing as its current categorization would indicate; "balanced" here refers to maintaining the 2-1 mesh refinement jump limit).  When cosmology tests get re-added back in they will need to be updated to shorten their times as well.

### DIFF
--- a/input/Adapt/adapt.incl
+++ b/input/Adapt/adapt.incl
@@ -12,13 +12,13 @@ include "input/Adapt/adapt_slope.incl"
 
 Stopping { 
 	 cycle = 10000; 
-	 time = 0.1;
+	 time = 0.05;
 }
 
 Boundary { type = "periodic"; }
 
 Testing {
-   time_final  = 0.10;
+   time_final  = 0.05;
    cycle_final = 0;
 }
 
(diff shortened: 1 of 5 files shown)


bwoshea:
@jobordner this seems pretty straightforward, but it seems that none of the CircleCI tests actually ran on this PR - do you have any idea why?  As I look at the actual files, I don't see any reason why they wouldn't run...

bwoshea (approved):

jobordner:
@bwoshea I'm not sure why they didn't run, maybe there were a couple of updates in quick succession which it didn't like.  I've pushed a small bugfix (see #98) to retrigger testing, and tests seem to be running now.

bwoshea:
@jobordner awesome, thank you!

gregbryan (approved):
Looks good to me -- thanks James!

jwise77 (approved):
Looks good. Thanks James!

PR #107, opened 2021-05-25
Auto-detect the number of cores in the build script

jwise77:
I got tired of stashing my changes to `build.sh` because I had `proc` a different value 😄 

By default, it will use all of the cores.  This can be overridden by setting either the proc variable in build.sh or the environment variable `CELLO_BUILD_NCORE`. I couldn't test whether this works on macOS

```
sysctl -n hw.ncpu
```

It'd be great if someone with macOS could test whether this command gives the number of cores.


mabruzzo (approved):
This is a fantastic change!

bwoshea (approved):

Issue #66, opened 2021-01-07
VL+CT testing updates


mabruzzo:
The VL+CT integrator tests should be ported to be more inline with the new testing framework introduced in PR #53. There are some ambiguities with how exactly the gold standard values should be treated.

Additionally, tests should be introduced to check the code when compiled in single precision mode (currently, the tests only run when compiled with double precision)

Some additional tests were discussed  discussed in the VL+CT PR (#62):

- It might be useful to modify the linear wave test (and other tests) so that we can easily change mesh sizes (currently, the sizes are hardcoded). This might make helpful for checking convergence rates
- We should move our gold standard acceptance values into external files (like a csv). Currently, these values are hardcoded into the test files. This may also help with testing the code in single precision mode
- Cleanup the machinery for computing L1 Error Norms. There's a lot of cruft in the current implementation that accrued because the design of the testing framework was not obvious at the time of writing.
- Refactor the shock tube test so that we can specify the initial left and right state in an input file
    - I feel like the best way to handle this is to introduce 2 new parameter groups to hold the left and right state (e.g. 'Initial:shock_tube:left_state' and 'Initial:shock_tube:right_state'). The parameters in each groups could easily be read into a vector of std::pairs or a mapping, and then passed directly passed to the initializer.
- Add some L_inf norm tests to for linear waves

PR #118, opened 2021-06-08
Introducing pup routines for the Value class

mabruzzo:
Up until now, the Value class has not had a working pup routine. This causes things to break in subtle and silent ways.

This issue has been known about for a while and was responsible for some complicated work-arounds in the `InitialValue` and `EnzoInitialBCenter` classes. 

I was recently reminded of this problem when I noticed some issues related to inflow boundaries in some of my science runs after restarts (these problems were primarily related to how I implemented the frame tracking scheme, but I also reproduced the issues on the main branch). This can also affect:
- `"mask"` type of refinement criterion
- general masks that can be specified in the Boundary section of the parameter file

My solution to this problem was to introduce the `Expression` class, which acts as a serializable pointer to the ``value_expr_`` field of the ``Param`` class. It's unfortunate that I had to introduce *another* layer to our evaluation of expressions, but it seems like the best solution given the following 2 constraints:

1. I went to great lengths to avoid duplicating the ``struct node_expr *`` pointers (since this wasn't done in the old implementation).
    - If we were willing to produce a single clone of the ``struct node_expr *`` hierarchy on each PE, that would dramatically simplify things. In that case, I would be tempted to:
        - move all methods that directly interact with the ``struct node_expr *`` hierarchy from the ``Param`` class to the ``Expression`` class and tweak the implementation of the ``Expression`` class so that it the class is the size of a pointer
        - have the ``Param`` class store an instance of the ``Expression`` class and introduce the ``Param::get_expression`` method where it returns a deepcopy/clone of the stored  ``Expression`` instance
2. All of the ``Parameter`` and ``Param`` classes' accessor methods mutate state
    - This creates some challenges after restart given that ``cello::parameters`` returns a ``const Parameter*``. In comparison to the first constraint, this is minor 
    - Directly addressing this constraint probably wouldn't change things too dramatically. It would probably just make things a little cleaner.
    - Given that the ``Parameter`` instance is ``readonly``, this may be something that we want to consider doing in the future anyway.

Some other minor changes include:
- I updated the copy/move constructors and assignment operators of the ``Value`` and ``ScalarExpr``. Let me know if you think I should revert these.
- I introduced the ``Parameters::leaf_parameter_names()`` method to help facilitate refactoring of ``InitialValue`` and ``EnzoInitialBCenter``. The refactoring dramatically simplifies the implementation and let's us avoid duplicating the global `Parameter` object during restart. I've actually wanted a method like this in the past to do things like:
    - specify expected levels of flux conservation for different quantities (in PR #108)
    - specify initial left and right shock tube states (to help address Issue #66)

One last side thought that might be worth considering for the future:
- It might be worth refactoring the ``Mask`` class and its subclasses so that they are stored in a ``std::unique_ptr`` instead of ``std::shared_ptr``. It seems like the only reason it's stored in a ``std::shared_ptr`` is for passing the Mask to `void ScalarExpr::evaluate(...) const`, and I think we could change the signature of `ScalarExpr::evaluate` to accept a `const Mask*` without having any worries about ownership. A benefit of this change is simpler PUP routines (charm++ has built-in support PUPing a ``std::unique_ptr``).


mabruzzo:
At this point, this probably shouldn't be merged in until after PR #97 since I'm pretty sure this will introduce conflicts

input/Gravity/method_gravity_cg.incl
```diff
@@ -90,11 +90,11 @@ Initial {
       total_energy  = [ 10.0 / (2.0/3.0 * 1.0),
                        (x)*(x) + (y)*(y) < 0.05,
                    1.0 / (2.0/3.0 * 0.1) ];
-       AP = 0.0;
+       #AP = 0.0;
```
brittonsmith, line 93: Why are these commented out?

mabruzzo:
I'm thinking that maybe we should table this PR in light of the fact that the new Checkpoint-Restart method isn't going to need to use PUP methods (and this PR is basically a giant workaround to get the PUP routine to work correctly).

PR #131, opened 2021-11-20
Update vlct test scripts

mabruzzo:
I've made 2 main changes to the python scripts used to run the VL+CT tests.

1. I made it possible to run those scripts from any directory. This should help a lot with executing these tests using `ctest` (in https://github.com/jobordner/enzo-e/pull/4 ). Longer term, we might want to refactor some of these tests.

2. I tweaked `run_dual_energy_shock_tube_test.py` to let you run the test in parallel with `charmrun` when it's not in your `PATH` (this was an oversight). This is done by looking at the `CHARM_HOME` directory.

### DIFF
--- a/input/vlct/run_HD_linear_wave_test.py
+++ b/input/vlct/run_HD_linear_wave_test.py
@@ -12,6 +12,7 @@
 #   2.) Check that the L1-norm error is the same for left propagating and right
 #       propagating sound waves
 
+import os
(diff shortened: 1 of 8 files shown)


gregbryan (approved):
Looks good to me!

jwise77 (approved):
Looks good, thanks!

PR #133, opened 2021-12-07
Quick Bugfix in RestartTest

mabruzzo:
It's come to my attention that the checkpoint-restart test for the VL+CT test is failing. I've confirmed that it's breaking for a very simple reason (I forgot to add the `Method:mhd_vlct:mhd_choice` parameter to the input file in PR #74).

But before I upload the commit solving this problem, I'm first going to make sure I've resolved a minor bug in `build.sh` that had let the script exit with an exit-code indicating success, even when some tests crashed. I believe, this prevented the bug from being reported by the Continuous Integration (in other words, I want to confirm that the Continuous Integration reports a failure once I've fixed the bug in `build.sh`, but not the VL+CT checkpoint bug).

**EDIT:** This is now ready for review. I also had to update the PUP routine for Grackle's `chemistry_data` struct. This should be extremely easy to review.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -172,7 +172,7 @@ printf "done\n"
 printf "done\n" >> $log
 
 # TESTS
-
+count_test_crashes=0
 if [ $target == "test" ]; then
 
     rm -f              test/STOP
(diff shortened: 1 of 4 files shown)


mabruzzo:
This is now ready to be reviewed!

gregbryan (approved):

mabruzzo:
This has 2 approvals. I'm going to merge this in.

PR #139, opened 2022-01-27
Replace `scons` build & test infrastructure with `cmake`

pgrete:
This replace the original PR to the fork https://github.com/jobordner/enzo-e/pull/4

Following the comments in the original PR, this PR will focus on getting the basic infrastructure in place, i.e.,
- build locally
- test locally and show results
- automate build & test in the CI infrastructure

The updated infrastructure will use simple markdown documents to show results (instead of php based ones) and replace flash/swf movies with gifs.

More detailed publication of tests results (e.g., using a `CDash` board) is left for a future update.

The PR is still WIP with following items left
- [x] create infrastructure to "process" test results (e.g., with `yt`)
- [ ] convert all tests
- [x] add basic `ctest` documentation
- [x] add documentation on how to add tests
- [x] remove old infrastructure and machine files
- [x] add more machine files (for machines of interest)
- [x] update CI scripts
- [x] dev doc: CHARM_PREPROC defs

PR #143, opened 2022-02-08
Multispecies Cosmology

WillHicks96:
Resubmitting this PR in a cleaner state

Changes are as follows:

1. Small bugfix in `enzo_SolveHydroEquations.cpp`, where `colindex` was initialized as NULL and never updated, which caused segfaults when doing color field advection.
2. Initializing chemistry fields in `enzo_EnzoInitialCosmology.cpp`
3. Explicitly calculating cosmology units in `EnzoMethodGrackle::setup_grackle_units()`. This is necessary because `setup_grackle_units` gets called in the constructor, which causes virtual functions `enzo_units->length()`, `time()`, etc. to reference the wrong version when doing cosmology runs
4. Added metallicity floor parameter
5. Added config file for building on Expanse

### DIFF
--- a/src/Enzo/enzo_EnzoConfig.cpp
+++ b/src/Enzo/enzo_EnzoConfig.cpp
@@ -234,6 +234,7 @@ EnzoConfig::EnzoConfig() throw ()
   method_grackle_chemistry(),
   method_grackle_use_cooling_timestep(false),
   method_grackle_radiation_redshift(-1.0),
+  method_grackle_metallicity_floor(0.0),
 #endif
   // EnzoMethodGravity
   method_gravity_grav_const(0.0),
@@ -619,6 +620,7 @@ void EnzoConfig::pup (PUP::er &p)
   if (method_grackle_use_grackle) {
     p  | method_grackle_use_cooling_timestep;
     p  | method_grackle_radiation_redshift;
+    p  | method_grackle_metallicity_floor;
     if (p.isUnpacking()) { method_grackle_chemistry = new chemistry_data; }
     p | *method_grackle_chemistry;
   } else {
@@ -1242,6 +1244,11 @@ void EnzoConfig::read_method_grackle_(Parameters * p)
     // for when not using cosmology - redshift of UVB
     method_grackle_radiation_redshift = p->value_float
       ("Method:grackle:radiation_redshift", -1.0);
+
+    // set a metallicity floor
+    method_grackle_metallicity_floor = p-> value_float
+      ("Method:grackle:metallicity_floor", 0.0);
+
 
     // Set Grackle parameters from parameter file
     method_grackle_chemistry->with_radiative_cooling = p->value_integer
--- a/src/Enzo/enzo_EnzoInitialCosmology.cpp
+++ b/src/Enzo/enzo_EnzoInitialCosmology.cpp
@@ -72,5 +72,29 @@ void EnzoInitialCosmology::enforce_block
   }
 
   block->initial_done();
-  
+
+#ifdef CONFIG_USE_GRACKLE
+  // initialize chemistry fields if doing multispecies 
+  if (enzo::config()->method_grackle_chemistry)
+  {
+    chemistry_data * grackle_chemistry =
+    enzo::config()->method_grackle_chemistry;
+ 
+    if ( (grackle_chemistry->primordial_chemistry > 0) || (grackle_chemistry->metal_cooling == 1))
+    {
+        Field field = block->data()->field();
+        EnzoBlock * enzo_block = enzo::block(block);
+        grackle_field_data grackle_fields_; 
+        code_units grackle_units_;
+ 
+        //create data struct to be fed into grackle
+        EnzoMethodGrackle::setup_grackle_units(enzo_block, & grackle_units_);        
+        EnzoMethodGrackle::setup_grackle_fields(enzo_block, & grackle_fields_);
+
+        //initialize density fields for various chemical species
+        EnzoMethodGrackle::update_grackle_density_fields(enzo_block, & grackle_fields_);
+    }
+  }
+#endif 
+
 }
--- a/src/Enzo/enzo_EnzoMethodGrackle.cpp
+++ b/src/Enzo/enzo_EnzoMethodGrackle.cpp
@@ -235,7 +235,38 @@ void EnzoMethodGrackle::setup_grackle_units (double current_time,
     enzo_float cosmo_dt = 0.0;
 
     EnzoPhysicsCosmology * cosmology = enzo::cosmology();
+    //~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+    //put into cosmology units
+    //
+
+    // Conversions are done explicitly here because this function is called in the constructor, which can lead to issues
+    // when calling virtual functions (e.g. enzo_units->length, density, etc.)
+    // Formulae are copied from src/Enzo/enzo_EnzoPhysicsCosmology.hpp
+
+    if (current_time == 0) {
+       current_time = cosmology->time_from_redshift(enzo::config()->physics_cosmology_initial_redshift);
+    }
+
+  double current_redshift = cosmology->redshift_from_time(current_time);
+
+  grackle_units->density_units  = 1.8788e-29*enzo::config()->physics_cosmology_omega_matter_now*
+      pow(enzo::config()->physics_cosmology_hubble_constant_now,2)*
+      pow(1 + current_redshift,3);
+  grackle_units->length_units   = cello::Mpc_cm*enzo::config()->physics_cosmology_comoving_box_size/
+      enzo::config()->physics_cosmology_hubble_constant_now/
+      (1.0 + current_redshift);
+  grackle_units->time_units     = 2.519445e17/sqrt(enzo::config()->physics_cosmology_omega_matter_now)/
+      enzo::config()->physics_cosmology_hubble_constant_now/
+      pow(1.0 + enzo::config()->physics_cosmology_initial_redshift,1.5);
+
+  grackle_units->velocity_units = 1.22475e7*enzo::config()->physics_cosmology_comoving_box_size*
+      sqrt(enzo::config()->physics_cosmology_omega_matter_now)*
+      sqrt(1.0 + enzo::config()->physics_cosmology_initial_redshift);
+
+    //
+    //~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    //
     cosmology->compute_expansion_factor(&cosmo_a, &cosmo_dt, current_time);
     grackle_units->a_units
          = 1.0 / (1.0 + enzo_config->physics_cosmology_initial_redshift);
@@ -383,6 +414,7 @@ void EnzoMethodGrackle::update_grackle_density_fields(
 
   double tiny_number = 1.0E-10;
 
+  const EnzoConfig * enzo_config = enzo::config();
   chemistry_data * grackle_chemistry =
     enzo::config()->method_grackle_chemistry;
 
@@ -411,6 +443,10 @@ void EnzoMethodGrackle::update_grackle_density_fields(
           grackle_fields->DII_density[i]   = grackle_fields->density[i] * tiny_number;
           grackle_fields->HDI_density[i]   = grackle_fields->density[i] * tiny_number;
         }
+
+       if (grackle_chemistry->metal_cooling == 1){
+          grackle_fields->metal_density[i] = grackle_fields->density[i] * enzo_config->method_grackle_metallicity_floor*cello::metallicity_solar;
+       }
 
       }
     }
@@ -557,6 +593,41 @@ double EnzoMethodGrackle::timestep ( Block * block ) throw()
   return dt * courant_;
 }
 
+//----------------------------------------------------------------------
+//
+#ifdef CONFIG_USE_GRACKLE
+void EnzoMethodGrackle::enforce_metallicity_floor(EnzoBlock * enzo_block) throw()
+{
+  // MUST have metal_density field tracked
+  Field field = enzo_block->data()->field();
+  enzo_float * density = (enzo_float*) field.values("density");
+  enzo_float * metal_density  = (enzo_float*) field.values("metal_density");
+
+  int gx,gy,gz;
+  field.ghost_depth (0,&gx,&gy,&gz);
+
+  int nx,ny,nz;
+  field.size (&nx,&ny,&nz);
+
+  int ngx = nx + 2*gx;
+  int ngy = ny + 2*gy;
+  int ngz = nz + 2*gz;
+
+  const EnzoConfig * enzo_config = enzo::config();
+  double Z_floor = enzo_config->method_grackle_metallicity_floor; 
+
+  for (int iz=0; iz<ngz; iz++){
+    for (int iy=0; iy<ngy; iy++){
+      for (int ix=0; ix<ngx; ix++){
+        int i = INDEX(ix,iy,iz,ngx,ngy);
+        double Z = metal_density[i] / density[i] / cello::metallicity_solar;
+        if (Z < Z_floor) metal_density[i] *= Z_floor/Z;
+      }
+    }
+  }     
+  return;
+}
+#endif
 //----------------------------------------------------------------------
 
 #ifdef CONFIG_USE_GRACKLE
--- a/src/Enzo/enzo_EnzoMethodGrackle.hpp
+++ b/src/Enzo/enzo_EnzoMethodGrackle.hpp
@@ -198,6 +198,8 @@ public: // interface
 			    "local_calculate_temperature");
   }
 
+  void enforce_metallicity_floor(EnzoBlock * enzo_block) throw();
+
 #endif
 
 protected: // methods
--- a/SConstruct
+++ b/SConstruct
@@ -262,6 +262,7 @@ elif (arch == "faraday_gnu"):  from faraday_gnu  import *
 elif (arch == "faraday_gnu_debug"):  from faraday_gnu_debug  import *
 elif (arch == "frontera_gcc"): from frontera_gcc import *
 elif (arch == "frontera_icc"): from frontera_icc import *
+elif (arch == "expanse_icc"):  from expanse_icc  import *
 elif (arch == "mf_gnu"):       from mf_gnu       import *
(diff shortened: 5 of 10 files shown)


src/Enzo/enzo_EnzoMethodGrackle.cpp
```diff
@@ -412,6 +444,10 @@ void EnzoMethodGrackle::update_grackle_density_fields(
           grackle_fields->HDI_density[i]   = grackle_fields->density[i] * tiny_number;
         }
 
+       if (grackle_chemistry->metal_cooling == 1){
+          grackle_fields->metal_density[i] = grackle_fields->density[i] * enzo_config->method_grackle_metallicity_floor*cello::metallicity_solar;
```
mabruzzo, line 448: Is there any reason why we aren't storing `metallicity_floor` as an attribute of the `EnzoMethodGrackle` class (instead of accessing it through `enzo_config->method_grackle_metallicity_floor`? 

If there isn't one, I'd request that you make that change (since this is the convention adopted throughout the rest of the codebase).

src/Enzo/enzo_EnzoMethodGrackle.cpp
```diff
@@ -557,6 +593,41 @@ double EnzoMethodGrackle::timestep ( Block * block ) throw()
...
+
+  int nx,ny,nz;
+  field.size (&nx,&ny,&nz);
+
+  int ngx = nx + 2*gx;
+  int ngy = ny + 2*gy;
+  int ngz = nz + 2*gz;
+
+  const EnzoConfig * enzo_config = enzo::config();
+  double Z_floor = enzo_config->method_grackle_metallicity_floor;
```
mabruzzo, line 617: same as my above comment about storing  `enzo_config->method_grackle_metallicity_floor`

src/Enzo/enzo_EnzoMethodGrackle.cpp
```diff
@@ -235,7 +235,38 @@ void EnzoMethodGrackle::setup_grackle_units (double current_time,
...
+
+    // Conversions are done explicitly here because this function is called in the constructor, which can lead to issues
+    // when calling virtual functions (e.g. enzo_units->length, density, etc.)
+    // Formulae are copied from src/Enzo/enzo_EnzoPhysicsCosmology.hpp
+
+    if (current_time == 0) {
+       current_time = cosmology->time_from_redshift(enzo::config()->physics_cosmology_initial_redshift);
+    }
+
+  double current_redshift = cosmology->redshift_from_time(current_time);
```
mabruzzo, line 251: Minor tweak: fix the indentation.
```suggestion
    double current_redshift = cosmology->redshift_from_time(current_time);
```

src/Enzo/enzo_EnzoMethodGrackle.cpp
```diff
@@ -235,7 +235,38 @@ void EnzoMethodGrackle::setup_grackle_units (double current_time,
...
+  grackle_units->length_units   = cello::Mpc_cm*enzo::config()->physics_cosmology_comoving_box_size/
+      enzo::config()->physics_cosmology_hubble_constant_now/
+      (1.0 + current_redshift);
+  grackle_units->time_units     = 2.519445e17/sqrt(enzo::config()->physics_cosmology_omega_matter_now)/
+      enzo::config()->physics_cosmology_hubble_constant_now/
+      pow(1.0 + enzo::config()->physics_cosmology_initial_redshift,1.5);
+
+  grackle_units->velocity_units = 1.22475e7*enzo::config()->physics_cosmology_comoving_box_size*
+      sqrt(enzo::config()->physics_cosmology_omega_matter_now)*
+      sqrt(1.0 + enzo::config()->physics_cosmology_initial_redshift);
```
mabruzzo, line 265: Can't we do the following:
```suggestion
    grackle_units->density_units  = 1.8788e-29*enzo::config()->physics_cosmology_omega_matter_now*
      pow(enzo::config()->physics_cosmology_hubble_constant_now,2)*
      pow(1 + current_redshift,3);
    grackle_units->length_units   = cosmology->length_units();
    grackle_units->time_units     = cosmology->time_units();
    grackle_units->velocity_units = cosmology->velocity_units();
```

mabruzzo: I'd be inclined to take this a step further and add a method to `EnzoPhysicsCosmology` called `EnzoPhysicsCosmology::density_units()` that we call here and in `EnzoPhysicsCosmology::mass_units()`. 

But if you don't want to handle this last part, just let me know and I'll make a new PR to do it.

src/Enzo/enzo_EnzoInitialCosmology.cpp
```diff
@@ -72,5 +72,29 @@ void EnzoInitialCosmology::enforce_block
...
+#ifdef CONFIG_USE_GRACKLE
+  // initialize chemistry fields if doing multispecies 
+  if (enzo::config()->method_grackle_chemistry)
+  {
+    chemistry_data * grackle_chemistry =
+    enzo::config()->method_grackle_chemistry;
+ 
+    if ( (grackle_chemistry->primordial_chemistry > 0) || (grackle_chemistry->metal_cooling == 1))
+    {
+        Field field = block->data()->field();
```
mabruzzo, line 85: I think you can drop this line
```suggestion
```

mabruzzo (changes requested):
This looks fantastic to me!

I just have a few minor requests/suggestions. Let me know if you disagree with any them

src/Enzo/enzo_EnzoConfig.cpp
```diff
@@ -1243,6 +1245,11 @@ void EnzoConfig::read_method_grackle_(Parameters * p)
     method_grackle_radiation_redshift = p->value_float
       ("Method:grackle:radiation_redshift", -1.0);
 
+    // set a metallicity floor
+    method_grackle_metallicity_floor = p-> value_float
```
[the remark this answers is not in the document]
WillHicks96: Sure thing!

src/Enzo/enzo_EnzoInitialCosmology.cpp
```diff
@@ -72,5 +72,29 @@ void EnzoInitialCosmology::enforce_block
...
+        EnzoBlock * enzo_block = enzo::block(block);
+        grackle_field_data grackle_fields_; 
+        code_units grackle_units_;
+ 
+        //create data struct to be fed into grackle
+        EnzoMethodGrackle::setup_grackle_units(enzo_block, & grackle_units_);        
+        EnzoMethodGrackle::setup_grackle_fields(enzo_block, & grackle_fields_);
```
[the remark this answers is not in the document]
WillHicks96: fixed

fearmayo:
Looks good to me @WillHicks96 . It would be good (probably necessary?) to also port across some of the cooling tests from enzo-dev (e.g. OneZoneFreefallTest and CoolingTest_Grackle). This would provide integration testing for this piece of code.

stefanarridge:
> Looks good to me @WillHicks96 . It would be good (probably necessary?) to also port across some of the cooling tests from enzo-dev (e.g. OneZoneFreefallTest and CoolingTest_Grackle). This would provide integration testing for this piece of code.

@fearmayo , is this what is being handled by PR #125 ?

fearmayo:
> > Looks good to me @WillHicks96 . It would be good (probably necessary?) to also port across some of the cooling tests from enzo-dev (e.g. OneZoneFreefallTest and CoolingTest_Grackle). This would provide integration testing for this piece of code.
> 
> @fearmayo , is this what is being handled by PR #125 ?


Possibly. @WillHicks96 can you check with @mabruzzo to see what the status of PR #125 is? If PR #125 is accepted then perhaps this PR could piggy back on top. I'm not sure that the tests introduced in PR #125 will test the new code introduced here though since this is specific to cosmology to a large degree 
- @WillHicks96 ?

WillHicks96:
@fearmayo Sure thing. I'm not super familiar with how the automated testing system works, but for testing this particular piece of code maybe it would be sufficient to add [input/test_cosmo-dd-multispecies.in](https://github.com/enzo-project/enzo-e/pull/143/files#diff-8884bd73c7e6eeebc04726e5922713c0cc6547534ac573bf6d0511d66b1e79d1) to the loop?

It looks like the test being called in [input/Grackle/grackle.incl](https://github.com/enzo-project/enzo-e/pull/125/files#diff-56eeede6cbd59eb13430c9e8e00abd824bc27bd335659b5f320feb111c666b34) from PR #125 is the same as CoolingTest_Grackle. Is this right @mabruzzo?

mabruzzo:
So I honestly forgot about PR #125. That PR is/was super ambitious... There were a couple of pieces to it:

- the `tools/gen_grackle_testing_file.py` script was used to generate the actual input file that would be used for running the test (it addresses the problem that we can't know the path to the data file ahead of time)
- the `tools/field_summary.py` script was a crude attempt to replicate some of the functionality of `enzo-dev`'s gold-standard testing. This script supported 2 subcommands:
    1. `measure` - the script measures field summary information from a snapshot and saves it to a text file
    2. `cmp` - the script measures field summary information from a snapshot and tries compare it against some reference summary statistics that were previously saved to disk.

Ideally, we would generate the reference data on the same machine that we are currently running the test on using an earlier version of Enzo-E and the same version of Grackle. But it was completely unclear to me how I would go about doing this with the Continuous Integration. Consequently, I tried to generate the reference statistics on my local machine and use those for future comparisons.

Unfortunately, there was disagreement in my results on the Continuous Integration. I can't remember if this was coming from using different Grackle versions or using different machines (it might not have been clear to me at the time). At the time, I set this aside because I found the problem really tedious to debug... 

I can take a look at that PR again over the next few days. My temptation would be to simplify the summary statistics that are checked. I think part of the problem I was encountering was that the location of the extrema would change - (maybe I would drop that for now). My basic thinking is having permissive tests may be better than nothing...

Alternatively, we can punt this down the line (and wait for more proper gold-standard testing)... But it's unclear when that will be (that's why I started #125 to begin with)

> It looks like the test being called in [input/Grackle/grackle.incl](https://github.com/enzo-project/enzo-e/pull/125/files#diff-56eeede6cbd59eb13430c9e8e00abd824bc27bd335659b5f320feb111c666b34) from PR https://github.com/enzo-project/enzo-e/pull/125 is the same as CoolingTest_Grackle. Is this right @mabruzzo?

That seems right. But I'm not totally sure... I was just recycling the file that Andrew Emerick had introduced... (My understanding is it's just solving a bunch of one-zone models in parallel)

mabruzzo:
Following up on this, I just finished up PR #125. 

If you wanted to make use of some of that machinery, I would recommend making an input file like that looks something like the following example:
```
include "input/Grackle/grackle.incl"

Physics {
    cosmology {
        comoving_box_size = 3.000000000000000;
        final_redshift = 3.000000000000000;
        hubble_constant_now = 0.7000000000000000;
        initial_redshift = 99.00000000000000;
        max_expansion_rate = 0.01500000000000000;
        omega_baryon_now = 0.04000000000000000;
        omega_cdm_now = 0.2600000000000000;
        omega_lambda_now = 0.7000000000000000;
        omega_matter_now = 0.3000000000000000;
    };
    list = [ "cosmology" ];
}

Initial { list = [ "grackle_test", "cosmology" ]; }

Method { list = [ "grackle", "null", "comoving_expansion" ]; }

Output {
    data {
        dir = ["GeneralGrackleCosmo-%02d", "cycle"];
        schedule {
            var = "cycle";
            list = [5];
        }
    }
}

Stopping { cycle = 5; }
```
(I'm not totally sure I've set up cosmology correctly, but the point is to avoid including a gravity solver since its not strictly deterministic - so that if results change, you know its related to a problem with cosmology & grackle). Then you could mimic the basic procedure for the grackle_general tests in `test/MethodGrackle/SConscript`. I'm happy to help out with that if it isn't obvious how to do it (I realize it's not the most obvious procedure to follow...).

Of course, this is only a suggestion. You could definitely also do something different.

src/Enzo/enzo_EnzoMethodGrackle.cpp
WillHicks96: Finally coming back to this -- sorry for the delay. I'm actually having trouble storing `metallicity_floor` as an attribute of `EnzoMethodGrackle` and accessing it here because this particular function is defined as a `static` method (so there's no `this` pointer). Do you know if there's any particular reason why or if it's necessary for these functions to be static?

mabruzzo: I believe that this method is static to make the method easier to call from problem initializers. But, I think it would be fine to make this function non-static. You would just need to replace to `EnzoMethodGrackle::update_grackle_density_fields` with `enzo::grackle_method()->update_grackle_density_fields` in `enzo_EnzoInitialCosmology.cpp` and `enzo_EnzoInitialIsolatedGalaxy.cpp`

Alternatively, it's probably fine to leave to leave this as written.

src/Enzo/enzo_EnzoMethodGrackle.hpp
```diff
@@ -198,6 +198,8 @@ class EnzoMethodGrackle : public Method {
 			    "local_calculate_temperature");
   }
 
+  void enforce_metallicity_floor(EnzoBlock * enzo_block) throw();
```
mabruzzo: Is this function actually used anywhere?

src/Enzo/enzo_EnzoMethodGrackle.cpp
WillHicks96: I went ahead and made this method, along with `setup_grackle_fields` and `delete_grackle_fields`, nonstatic, and called them using `enzo::grackle_method()`, since this is consistent with how other Grackle methods are being called externally elsewhere in the code (e.g. `calculate_temperature` in `EnzoComputeTemperature`)

src/Enzo/enzo_EnzoMethodGrackle.cpp
```diff
@@ -235,7 +235,38 @@ void EnzoMethodGrackle::setup_grackle_units (double current_time,
...
 
     EnzoPhysicsCosmology * cosmology = enzo::cosmology();
+    //~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+    //put into cosmology units
+    //
+
+    // Conversions are done explicitly here because this function is called in the constructor, which can lead to issues
+    // when calling virtual functions (e.g. enzo_units->length, density, etc.)
+    // Formulae are copied from src/Enzo/enzo_EnzoPhysicsCosmology.hpp
```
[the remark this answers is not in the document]
WillHicks96: Thanks for clearing this up! I've updated this comment

src/Enzo/enzo_EnzoMethodGrackle.hpp
WillHicks96: I must have forgotten to copy over the line calling this function when preparing the branch for this PR. Added that back in. It's now called in `compute_` just after `local_solve_chemistry`

peeples:
Looks like this has a few requested changes and then is ready to merge, especially now that #139 is merged -- yes, @WillHicks96 ?

WillHicks96:
@peeples Yes, that's right. I've made all of the changes requested by @mabruzzo. 
I'm looking into the cooling tests requested by @fearmayo now that I don't have any excuses to put it off since PR #139 is merged.

mabruzzo:
Doesn't the `EnzoInitialGrackleTest` already effectively implement the one-zone free-fall test?

WillHicks96:
@mabruzzo I think so -- they're very similar. The only functional difference between the two that I can see is that in the one-zone free-fall test, the density field is updated over time via some analytic expression, which would require a method to to be added in Enzo-E to implement (this is the last thing that was on my to-do list for this PR).

mabruzzo:
So I took a look at your changes and it looks like the Grackle test you added to the automated testing suite, `CoolingTest_Grackle` is extremely similar to one of the tests (`grackle_cooling_dt`) that I added in PR #125 (that PR just got merged into main).  My version of the test let's us remove `CloudyData_UVB=HM2012.h5` from the repository (the test will use the version of that file downloaded with Grackle). 

So here's what I'm going to recommend:

1. merge the changes from main into your branch
2. remove `CoolingTest_Grackle`  from `tests/CMakeLists.txt`
3. Remove `CloudyData_UVB=HM2012.h5` from the repository. As I mentioned above, this file was previously included in the repository and was removed in PR #1. For that reason, I don't really think we want to track that file in the Enzo-E repository over the long-term.

Then I think maybe we just merge your branch into main. We can talk about additional tests next week

WillHicks96:
@mabruzzo Sounds good to me. I'll remove the one zone freefall test initializer for now too to avoid cluttering the rep with unused files

WillHicks96:
@mabruzzo Okay, made the changes you suggested!

mabruzzo (approved):

peeples:
@fearmayo It looks like the changes you requested have been addressed here; can you take a look so this can be merged? Thanks!

fearmayo (approved):

mabruzzo:
When I tried to resolve the conflict I broke the code. We need to:
1. Replace all 3 occurrences of `cello::metallicity_solar` with `enzo_constants::metalicity_solar` in `enzo_EnzoMethodGrackle.cpp`
2. add the lines
  ```c++
    // Solar metallicity
    const double metallicity_solar = 0.012;
  ```
 to `enzo_constants.hpp`
3. Remove `refresh_smooth->set_solver_id(index());` from `enzo_EnzoSolverJacobi.cpp` (I'm not sure how that got reintroduced...)

And then everything seems to compile. I don't seem to have permission to manually push changes to this branch. Maybe it's because Will's fork is based off of James's fork? Maybe someone else can try?

fearmayo:
Hi Will, 



> When I tried to resolve the conflict I broke the code. We need to:
> 
>     1. Replace all 3 occurrences of `cello::metallicity_solar` with `enzo_constants::metalicity_solar` in `enzo_EnzoMethodGrackle.cpp`
> 
>     2. add the following lines to `enzo_constants.hpp`:
> 
> 
> ```c++
>   // Solar metallicity
>   const double metallicity_solar = 0.012;
> ```
> 
>     3. Remove `refresh_smooth->set_solver_id(index());` from `enzo_EnzoSolverJacobi.cpp` (I'm not sure how that got reintroduced...)
> 
> 
> And then everything seems to compile. I don't seem to have permission to manually push changes to this branch. Maybe it's because Will's fork is based off of James's fork? Maybe someone else can try?

I tried too and I was able to push either. For me the code compiles too.  I didn't need to do step `3` above though - I don't see that conflict in my clone. I thought Stefan's PR fixed that issue?

WillHicks96:
Just made the changes. I need to break off from James' branch at some point...

Issue #115, opened 2021-06-02
Change documentation references to "value" boundaries to refer to "inflow" boundaries 


mabruzzo:
Currently the Reference Guide on the website for the Enzo-E/Cello Parameters talks about `"value"` type of boundary condition. These references to `"value"` boundaries should all be changed so that they instead refer to the `"inflow"` type of boundary condition (to reflect the fact Enzo-E/Cello recognizes the `"inflow"` type and not the `"value"` type).

 I believe that this confusion relates to the fact that `"inflow"` boundary is implemented with the `BoundaryValue` class

PR #144, opened 2022-02-14
Replace value with inflow for Boundary Parameter #115

buketbenek:
If I understand the issue correctly, I replaced "value" parameter with "inflow" parameter for the Boundary in Documentation. "make html" worked fine.


mabruzzo (approved):
Looks great to me! (You completely addressed the issue!)

stefanarridge (approved):
All looks fine to me!

PR #150, opened 2022-02-18
Gravity timestep

stefanarridge:
This PR changes the way that the timestep is calculated in `EnzoMethodGravity::timestep_`. A new parameter, `Method:gravity:dt_max` is introduced, which sets the value of the `dt_max_` attribute of `EnzoMethodGravity`, and the timestep is calculated as folows:

1.  `mean_cell_width` is calculated by taking the geometric mean of cell widths along all dimensional axes.
2. `epsilon` is defined as `mean_cell_width / (dt_max_ * dt_max_)`
3. The magnitude of the square acceleration in all cells is calculated, with `a_mag_2_max` set to the maximum value across all cells.
4. Set `a_mag_max` to be the square root of `a_mag_2_max`.
5. The timestep is calculated as `sqrt(mean_cell_width / a_mag_max + epsilon)`

This definition of the timestep has the property that its value is independent of how the acceleration vectors are oriented relative to the mesh (unlike what is implemented now). If the acceleration is zero everywhere, such as at the first timestep, then the timestep is equal to `dt_max_`. Another advantage is that (as of commit cbfb5a4), the `timestep_` function no longer does a square root operation for every cell in a block.

Currently, the default value of `dt_max_` is `1.0e10`; I'm not completely sure if that is a sensible default value. It would be up to the user to set a suitable value depending on the problem and on the units used.


gregbryan (approved):
This looks sensible to me.

fearmayo (approved):

PR #153, opened 2022-03-07
Address compiler warning - stop implicit cast from ``const char*`` to ``char*``

mabruzzo:
As the title implies, this avoids several unnecessary casts from ``const char*`` to ``char*`` (which previously raised compiler warnings)


gregbryan (approved):

stefanarridge (approved):
Looks fine to me.

PR #158, opened 2022-03-30
Method helper functions

stefanarridge:
This PR implements the following changes:

- Adds member functions to the Method class called `method_exists` and `method_precedes`, which both return a `bool`.
- `method_exists` takes the name of a method as an argument, and returns whether or not that method is being used for the current problem being run.
- `method_precedes` takes two names as arguments. It returns true if both methods are being used and the first method precedes the last in the list of methods (provided in parameter file), and false otherwise.
- `EnzoInitialMergeStarsTest::enforce_block` checks whether `merge_stars` and `pm_update` methods are being used.
- `EnzoMethodPmDeposit::compute` and `EnzoMethodGravity::compute` check (at the initial cycle) that `pm_deposit` precedes `gravity', and by extension, ensure that if one method is used, so is the other.
- Edited `input/Gravity/method_gravity_cg.incl` to include `pm_deposit` in `Method:list`.
- Removed `input/Gravity/method_gravity_mg.incl`, as it wasn't being used in the tests.


src/Cello/problem_Problem.hpp
```diff
@@ -137,6 +137,14 @@ class Problem : public PUP::able
   /// Return the named method object if present
   Method * method (std::string name) const throw();
 
+  // Return whether a method object with given name exists for this problem
+  bool method_exists (std::string name) const throw();
+
+  // Returns true if method objects with both given names exist for this problem,
+  // and the method called "name1" precedes the method called "name2" in the
+  // method list. Returns false otherwise.
```
mabruzzo, line 145: You should specify the behavior in edge-case where you have more than one instances of the Method called  "name1" and the method called "name2" occurs between them. I'm pretty sure the function returns `true`.

This edge-case is not pathological, there are certain methods (like the flux correction) that are designed to support multiple executions per block per cycle.

src/Cello/problem_Problem.hpp
```diff
@@ -137,6 +137,14 @@ class Problem : public PUP::able
   /// Return the named method object if present
   Method * method (std::string name) const throw();
 
+  // Return whether a method object with given name exists for this problem
+  bool method_exists (std::string name) const throw();
```
mabruzzo, line 141: You might want to consider passing a `const std::string&` instead, but it's not a big deal (since you only really use this while setting things up)

src/Cello/problem_Problem.hpp
```diff
@@ -137,6 +137,14 @@ class Problem : public PUP::able
...
+
+  // Returns true if method objects with both given names exist for this problem,
+  // and the method called "name1" precedes the method called "name2" in the
+  // method list. Returns false otherwise.
+  bool method_precedes(std::string name1, std::string name2) const throw();
```
mabruzzo, line 146: Same advice as above: you might want to consider passing a `const std::string&` instead, but it's not a big deal (since you only really use this while setting things up)

src/Cello/problem_Problem.cpp
```diff
@@ -832,6 +832,41 @@ Method * Problem::method (std::string name) const throw()
...
+    }
+    if (method_list_[i]->name() == name2) {
+      ind2 = i;
+      method_2_found = true;
+    }
+    if (method_1_found && method_2_found) break;
+  }
+
+  if (!(method_1_found && method_2_found)) return false;
+  else return ind1 < ind2;
```
mabruzzo, line 865:
```suggestion
  return method_1_found & method_2_found & (ind1 < ind2);
```
Feel free to ignore this.

mabruzzo (approved):
Looks great to me. I would just tweak the docstring to address the behavior for the edge case I mentioned

fearmayo (approved):

src/Cello/problem_Problem.hpp
stefanarridge: Thanks for pointing this out. I hadn't considered that methods might be executed more than once per cycle. 

I suppose what I really want to check is that "name2" does not precede any occurrence of "name1". I suppose then I should probably rename the function to `method_does_not_precede`, and then switch around the arguments in the calls to this function, but that would make it a bit more verbose. I think the intended meaning of `method_precede` is still clear enough.

stefanarridge:
Hi Molly,

Could you roll back this merge commit please? I would do it myself bit I'm
currently on my phone. It wasn't quite ready yet. Matthew approved it but
pointed out a mistake that needed to be fixed. I was planning to address it
tomorrow and then merge into main.

Kind regards,
Stefan

PR #160, opened 2022-04-04
Configured website to support doxygen docstrings with Breathe extension

mabruzzo:
This should simplify the process of writing new development/design documentation for the website that describes how to use/extend the codebase. It should also reduce the burden of keeping this documentation up to date with changes made in the codebase.

A minimal working example has been included to make sure that this works properly.


mabruzzo:
@jobordner or @brittonsmith can you confirm the read-the-docs is currently configured to using the web-interface? (The alternative is a [configuration file](https://docs.readthedocs.io/en/stable/guides/reproducible-builds.html#using-a-configuration-file) ).

For this PR to work, we need to use pip to install the `breathe` module (and I think we need a version of sphinx newer than 3.0).

mabruzzo:
I addressed the Read the Docs problem by creating a pip "requirements.txt" file to force the breathe module to be installed.

There are 2 ways to inform Read the Docs of the location of the existence/location of this file:
1. Modify the appropriate field on the `Admin` > `Advanced` page of the website interface.
2. Specify the file in a `.readthedocs.yaml` file that is located in the root directory (that overrides all the settings on website interface)

I originally went with the first approach, but this caused the Read the Docs builds to break for other PRs since they did not have the new `requirements.txt` file (In particular it briefly broke the build test in PR #139). As a result, I went with the second approach (since that doesn't break the other PR build tests). 

We can always swap approaches in the future, once the `requirements.txt` file makes its way into the other branches.

gregbryan (approved):

brittonsmith (approved):

peeples:
@mabruzzo go ahead and merge if you think this is ready!

PR #145, opened 2022-02-14
VL+CT Riemann Kernel

mabruzzo:
This PR depends on PR #139 (introduction of the CMake build-system).

As I describe below in more detail in [this comment](https://github.com/enzo-project/enzo-e/pull/145#issuecomment-1087722589), 
- the file changes introduced by this PR are actually 3056 insertions & 1296 deletions that affect 37 files (about half of what GitHub shows)
- the failing test in the CI can safely be ignored
In that comment, I also recommend how this PR can be reviewed (so that reviewers don't have to wait for PR#139 to be merged to be able to start reviewing this PR)

 
# Summary

This PR largely focuses on refactoring the Riemann Solver. In service of this refactor, it also involves some related changes to the implementation of `EnzoEFltArrayMap` & to the build-system.

The changes to the implementation of `EnzoEFltArrayMap` are described in the following spoiler tag:

<details>
  <summary>EnzoEFltArrayMap Changes</summary>

  * `EnzoEFltArrayMap` now internally stores its array data in a new template class called `CArrCollec` (which is implemented in the Array component of the Cello layer).

  * `CArrCollec` is effectively a union that supports 2 different approaches forstoring an ordered collection of 3D `CelloArray`s that share a fixed shape:

     1. A vector of `CelloArray`s (equivalent to an array of pointers)
     2. A single contiguous 4D `CelloArray`

  * Each approach has a nearly identical interface except that the latter can provides access to the underlying array (the way they are constructed is also somewhat different)

  * In the long-term, we probably just want to use a single backing implementation, but in the short-term, this substantially simplifies things. (This may merit further discussion - there might actually be an argument for supporting both approaches)

</details>

Changes to the Riemann Solvers include:

  * moving the Riemann Solver source/header files into the subdirectory `src/Enzo/EnzoRiemann`, and modifying the build system to compile the Riemann Solvers in a library separate from the rest of the Enzo-layer. The change in code organization follows conventions that were previously discussed (but never implemented). Compilation as a separate library **substantially** improves the speed with which you can develop the Riemann Solver code. (The template-heavy nature of this code previously would cause a rebuild of the entire Enzo-layer to be triggered whenever **any** implementation detail changed). This led to some changes to the PUP routines that appear a little messy, but can be removed once we move to our new restart strategy.
  * I refactored the way the Riemann Solvers are implemented to use a kernel-based approach that should be *very* compatible with porting to Kokkos.
[...]

PR #161, opened 2022-04-06
Make `EnzoEFltArrayMap` cheap to copy`

mabruzzo:
This PR makes the `EnzoEFltArrayMap` a lot cheaper to copy (this is a PR I've been working slowly for a long time).

These changes replace the classes used to represent data for 2 members the `EnzoEFltArrayMap` class: `str_index_map_` and `arrays_`. Previously, both members used standard library containers (`std::map` and `std::vector`). Copies of the standard library containers involve heap allocations and require that all of the contents of the containers are copied.

The changes introduced in this PR take advantage of the fact that the contents of `EnzoEFltArrayMap::str_index_map_` and `EnzoEFltArrayMap::arrays_` are not allowed to change* after `EnzoEFltArrayMap` is constructed. The new containers store their contents in memory whose lifetimes are managed by `std::shared_ptr`. 
- When we make a copy of these containers, the `std::shared_ptr` is copied instead of heap-allocating memory for the new container and copying each item between the containers.
- I elected not use `std::shared_ptr<std::map>` and `std::shared_ptr<std::vector>` because this would introduce an extra pointer indirection each time the container's contents are accessed. Additionally, the replacement for `std::map` has room for additional optimizations that could make it faster (and it could be reused in other areas of the codebase like `Performance` or possibly `FieldDescr`)

More details are provided in the commit messages. A few minor implementation choices were made to reduce conflicts with changes introduced in PR #145 

*As a point of clarification: the elements stored within `EnzoEFltArrayMap::arrays_`. The `CelloArray` instances themselves are not allowed to be mutated (e.g. you can't swap it with another `CelloArray`), but since they are smart pointers, the associated array data can still be mutated.


gregbryan (approved):
Looks very nice!

peeples:
@mabruzzo looks like this is good to go, with that one SCons thing to resolve

PR #164, opened 2022-04-12
Method helper functions 

stefanarridge:
This is essentially a repeat of PR #158, but with changes applied following @mabruzzo 's review:

1 - `Problem::method_exists` and `Problem::method_precedes` now take references to `const std::string` as arguments rather than just plain `std::string`s. 
2 - `Problem::method_precedes` returns true only if both methods appear once (and only once) in the list of methods, and if the first method precedes the second.

I ran clang-format on the source and header files I edited before committing my changes, which unfortunately means that there are small, stylistic changes to a several hundred lines of code in these files. The only parts that really need to be reviewed are the `method_precedes` and `method_exists` functions, and the calls to these functions.


mabruzzo:
In the future, I think it would be very useful to use clang-format (and add it to the CI). 

It has actually been on my todo list to work on finding customizing clang-format parameters to try to approximate the existing style. If you wanted to try to do that, you should definitely go for it. But, I would strongly suggest that you make a PR dedicated to that change (we would need to get buy-in from everybody on such a change - since it will be a major inconvenience for everybody with existing branches. We might need to introduce such a change in phases).

With that said, until we do something like that, I would **strongly** advise against using clang-format in isolation like this. Just so that you are aware for the future, doing this causes 2 problems:
- The primary issue this causes is merge conflicts. For example, if anybody else has made any changes ``Cello/problem_Problem.hpp`` or ``Cello/problem_Problem.cpp`` in their own local branches, they are now likely to encounter merge conflicts. These conflicts are particularly annoying to resolve since you need to need to examine each conflict to determine whether it arose from a stylistic change or if it arose from incompatible code changes or both. (This isn't just a hypothetical scenario, I personally have a custom/local branch that I'm using for some science runs that will be affected by this if I ever try to upstream some of the changes or update the branch with more recent changes).
- A much less significant issue is that this makes tracking the history for a particular line of code much more difficult.

If these changes had not already accidentally been merged into the main branch and then reverted, I would ask you to change this. But, at this point, it probably doesn't make a difference.

fearmayo:
> 

I see - thanks for this Matthew. I've two comments questions then in that case. 
1. Do you advise that we reject this PR because the clang-format tool has caused a huge amount of code change somewhat unrelated to the PR? I think that is what you are advising? Maybe then Stefan can resubmit the PR without the clang-formatting - to be fair to Stefan I think he was just looking for a tool that easily eliminate trailing whitespaces without making the wholesale changes we see here. 
2. It may be helpful if you lead a discussion on the Monday of the workshop on a preferred code formatting standard and we can all then agree on that going forward.

stefanarridge:
Yes it's probably best if I just open another PR, with only the meaningful changes. It's not a huge amount of work so I should be able to do it today.

PR #166, opened 2022-04-16
Bugfix: add nonexistent field to group

mabruzzo:
This bug arose because of an annoying property of `std::map`:
- In this scenario, `field_index` is an instance of `std::map<std::string,int>` that maps field names to field indices.
- While adding fields to groups, based on an input configuration file, the old code checked for the index associated with a field name, `field`, using `field_index[field]`.
- Unfortunately, if `field` is not already contained by `field_index`, a new key-value pair is inserted into `field_index` (This is required to support the syntax like `field_index["density"] = 3;`).
- As a result, when users tried to add previously undeclared fields to a group, a different field was added to the group instead.

An alternative fix would be to replace `field_index[field]` with `field_index.at(field)`. In cases where `field_index` does not already contain field, an exception would be raised. However, my chosen solution provides a more detailed error message that tells users what they did wrong.

While doing this, I also noticed that it was unnecessary for `field_index` to be an attribute of `Config`. The map is constructed and used inside of `Config::read_field_` and then it is never touched again. With that in mind, I removed `field_index` from `Config`'s attributes and made it into a local variable.

### DIFF
--- a/input/Checkpoint/checkpoint_grackle.in
+++ b/input/Checkpoint/checkpoint_grackle.in
@@ -73,8 +73,12 @@
    # but not everything
     color {
        field_list = [
-              "HI_density",
-              "HII_density",
+# because the entries for the first 2 fields were purposefully commented out
+# from Field::list, we need to comment them out here too (you can't add a
+# non-existant field to a group). This is okay: when EnzoMethodGrackle creates
+# these missing fields, it will automatically add them to the "color" group
+#              "HI_density",
+#              "HII_density",
               "HM_density",
               "HeI_density",
               "HeII_density",
--- a/input/PPML/ppml.incl
+++ b/input/PPML/ppml.incl
@@ -52,7 +52,6 @@
       }
 
    }
- include "input/flux_correct.incl"
 
    Initial {
 
--- a/src/Cello/parameters_Config.cpp
+++ b/src/Cello/parameters_Config.cpp
@@ -65,7 +65,6 @@ void Config::pup (PUP::er &p)
 
   p | num_fields;
   p | field_list;
-  p | field_index;
   p | field_alignment;
   PUParray(p,field_centering,3);
   PUParray(p,field_ghost_depth,3);
@@ -510,6 +509,8 @@ void Config::read_field_ (Parameters * p) throw()
 
   field_list.resize(num_fields);
 
+  std::map<std::string, int> field_index;
+
   for (int i=0; i<num_fields; i++) {
     field_list[i] = p->list_value_string(i, "Field:list");
     field_index[field_list[i]] = i;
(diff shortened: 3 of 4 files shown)


gregbryan (approved):
Looks good to me!

gregbryan:
Oops -- I see there is a test failure related to the grackle restart. That doesn't seem likely to be related to this PR.

stefanarridge:
Looks okay to me as far as I can see, will be happy to approve when the tests all pass.

stefanarridge (commented):
Looks okay to me as far as I can see, will be happy to approve when the tests all pass.

mabruzzo:
It turns out that the failing tests were related to config files that attempted to add non-existent fields to field-groups (so the new code correctly made the tests fail).

It turns out that the results of the affected test problems weren't actually affected by these issues (but the errors signaled potential issues for that would affect slightly more sophisticated setups). I provide a basic summary of the affected test problems below (more details are provided in the message of the commit that addressed this issue):  
-  Prior to this PR, the error in the input file for the Checkpoint-Restart test involving Grackle effectively caused "density" field to be flagged as "color" field. 
    - This issue would only cause issues with advection, and this test problem does not involve a hydro-solver. (As an aside, it is my intention to make sure we have a test problem involving Grackle and a hydro-solver, but I'm pursuing that elsewhere)
    - Even if this test did include a hydro solver, the test would still pass (since the Checkpoint-Restart test only cares that results are consistent in runs with and without a restart - it didn't care about the correctness of the results).
- The issue relating to the Ppml tests had to do with the improper application of flux flux corrections. This issue stems from the fact that the Ppml solver expects fields to have different names from from the fields used in the Ppm and VL+CT solvers. Basically, the config file tried to added hydro fields, (not used by the Ppml solver) to field-groups required by the flux correction machinery. 
    - As a consequence Flux Corrections were not really working properly
    - This didn't really matter since the affected Ppml tests are all unigrid (so I think `MethodFluxCorrect` doesn't actually do anything). To fix this, I just stopped applying flux corrections.

gregbryan:
Nice! Great to see that this fix uncovered some existing test issues. I looked over the new changes and they look very sensible. Reapprove!

stefanarridge (approved):
Looks good to me, good work on tracking down the reason for the Restart tests failing.

It might be worth creating an issue documenting the problems you encountered with the Restart tests. It could be useful for PR #139. In particular, for Alastair from ICHEC, who is working on converting these tests to the `ctest` infrastructure.

mabruzzo:
> It might be worth creating an issue documenting the problems you encountered with the Restart tests. It could be useful for PR #139.

Gotcha, I'll open a new issue to mention the possibility that there could be an issue. Frankly, this should not be a problem at all since this issue had nothing to do with checkpoint-restart functionality at all (the new error-checking will simply causes the simulation to abort with an error message while parsing the configuration file). As I mentioned before, the changes to the config files won't actually affect the simulations outputs anyway. The only way I could imagine this being a problem if Alastair pulls source code changes but continues to use the old config files.

> In particular, for Alastair from ICHEC, who is working on converting these tests to the `ctest` infrastructure.

If he has any issues/questions about this, you should point him in my direction. I wrote the python scripts to run several of the Checkpoint-Restart tests (other than the Ppm and Ppml tests). The scripts are a little more complex than they needed to be (given that I was unsure what the future of the testing framework would look like). It should actually be very straight-forward to port over these tests.

PR #171, opened 2022-04-27
"merge_stars" updated and renamed to "merge_sinks"

stefanarridge:
This PR introduces the following changes:

- `"merge_stars"` method is renamed to `"merge_sinks"`, and merges `"sink"` particles rather than `"star"` particles.
- `"merge_stars_test"` initializer is renamed to `"merge_sinks_test"`.
- The `"merge_sinks"` method now only calls the Friends-of-Friends algorithm if there is more than one sink particle in the block, which includes sink particles copied from neighbouring blocks.
- In the `"merge_sinks"` method, particles which have been "merged into" another particle, and therefore need to be deleted, now have their positions shifted by three times the block width, rather than two times. This was necessary to fix a bug which was not picked up by the tests previously.
- A new pair of tests of the `"merge_sinks"` method have been added to the test suite (one serial, one parallel). In these tests, a sphere of sink particles are initialised, with the center of the sphere being at the center of the domain, with the sink particles also having a velocity directed towards the center of the sphere. There is no additional "drift velocity" added to the particles' velocities, so that the center of the sphere remains stationary. These are in addition to the previous pair of tests, in which the center of the sphere is in the middle of a block, and a drift velocity is added to all the particles' velocities.
- The documentation has been updated accordingly.


doc/source/user/problem_method.rst
```diff
@@ -574,38 +574,37 @@ In addition, its 'lifetime' attribute is set to be the maximum of the group,
...
-Currently this will only run in unigrid mode. This is because when a block
-calls this method, it assumes that all neighbouring blocks have the same size,
-or equivalently, on the same refinement level.
-For this reason, there is a check in the constructor of EnzoMethodMergeStars
-for whether ``"Adapt: max_level"`` is equal to zero. In the future, we plan to
+Currently this will only run in unigrid mode. This is because this method
+will only work correctly if all blocks containing sink particles are of the
+same size, or equivalently, on the same refinement level.
+For this reason, there is a check in the constructor of EnzoMethodMergeSinks
+for whether ``"Adapt: max_level"`` is equal to zero. In future, we plan to
```
jwise77, line 589: "In the future" was correct for American English :)

jwise77:
This all looks good. I have a few questions that pertain to the merging that weren't necessarily changed in this PR but introduced in the original sink PR that I didn't review.

- For my own reference, how you specify to copy the particles from the neighboring blocks in the refresh? I couldn't figure it out.
- When you copy the particles from neighboring blocks, would there be a case where it can be merged into two sinks? That is, the copied sink particle (from the neighboring block) is merged into a massive sink on the current block, while the original sink particle is merged into a sink on the same block as it. I hope that makes sense.
- Also, do you delete the particles on the neighboring blocks after they're merged?

stefanarridge:
In response to your questions @jwise77:

> For my own reference, how you specify to copy the particles from the neighboring blocks in the refresh? I couldn't figure it out.

This is done in lines 39-44 of `src/Enzo/enzo_EnzoMethodMergeSinks.cpp`:

```
// Refresh copies all sink particles from neighbouring blocks
  cello::simulation()->refresh_set_name(ir_post_,name());
  Refresh * refresh = cello::refresh(ir_post_);
  ParticleDescr * particle_descr = cello::particle_descr();
  refresh->add_particle(particle_descr->type_index("sink"));
  refresh->set_particles_are_copied(true);
```
The code which actually implements the copying is in files in `src/Cello/`, probably best to start from `src/Cello/control_refresh.cpp`.

>When you copy the particles from neighboring blocks, would there be a case where it can be merged into two sinks? That is, the copied sink particle (from the neighboring block) is merged into a massive sink on the current block, while the original sink particle is merged into a sink on the same block as it. I hope that makes sense.

No, this cannot happen. The best way to think of the merging procedure is as follows:
- Each block copies all sink particles from neighbouring blocks.
- Run the Friends-of-Friends algorithm, which assigns each particle to one (and only one) FoF group. If a particle is within the merging radius of two other particles, then all three particles are in the same FoF group.
- If a FoF group contains more than one particle, these particles are all merged together. That is to say, the particles are replaced by one particle, whose position is equal to the centre of mass of the particles in the FoF group. In practice, I change the properties of one of the particles, and delete the rest.
- Out of the particles which remain (after looping over all FoF groups), particles which are out of bounds of the block are deleted. I am slightly concerned about this part, as it assumes that neighbouring blocks will always "agree" on which block a particle should belong to. There may be a situation where the centre-of-mass of a FoF group could be very close to a block boundary, and due to rounding errors, the blocks either side of the boundary may get slightly different values for the centre-of-mass, and disagree on which side of the boundary the centre-of-mass is on. I hope that this is a pathological case which won't happen in practice, it certainly does not seem to have come about in my tests.

There is another issue, however: this only works if every FoF group only contains particles which originate from neighbouring blocks. See my sketch here for a situation where this isn't the case:


Block B thinks that some particles originating from A should be merged with particles originating from C, but since blocks only copy particles from neighbouring blocks, A cannot see the particles from C, and vice versa. For this reason, I put in a check that if this situation ever happens, the program exits with an error message, advising the user that they might need to use larger blocks. Again, I would think that this is a bit of a pathological case: in practice, I imagine the FoF groups will nearly always contain a small number of particles (probably 1 or 2 most of the time) and I don't think having a long chain of particles is something that would naturally arise.

>Also, do you delete the particles on the neighboring blocks after they're merged?

I think I already answered this in the previous point, but to clarify, the only particles which remain on the block after the merging procedure are the "post-merger" particles which are "in bounds" of the block. Note that it is possible to merge two particles from neighbouring blocks and get a particle which is on the block, as shown in this sketch:



Anyway, I hope that clears things up, let me know if you have any more questions.

jwise77 (approved):
Thanks so much for answering my questions in great detail and providing some sketches! This all clears things up, and I can approve the PR now.

fearmayo (approved):

PR #174, opened 2022-05-02
`EnzoPhysicsFluidProps`

mabruzzo:
## Overview

I have refactored the Enzo-layer to make use of the `EnzoPhysicsFluidProps` class. The main idea behind this class is to store fluid properties (that may need to be accessed by multiple `Method` or `Compute` objects) in a central location. Currently the class stores:
- dual energy formalism configuration
- quantity floors (currently just the density, pressure, temperature, and metal-mass-fraction floors are tracked here - in a future PR we'll probably add support for tracking a more general passive scalar floor)
- the mean molecular weight
- the adiabatic index

## Motivation
This PR was motivated by a couple of factors:
1. Previously, functions/Methods looked for the parameters now tracked by 
the `EnzoPhysicsFluidProps` class in a number of different locations throughout the codebase (e.g. some checked attributes of `EnzoConfig` and others were checking some global variables that were defined in `EnzoBlock`)
2. Most of these parameters were defined within the `ppm` solver section of the parameter file. Previously we had to either duplicate these parameters in the `mhd_vlct` solver section of the parameter file (like the density floor, pressure floor, and decision whether to use the dual-energy formalism). This created confusion about where other Method objects (like `EnzoMethodComovingExpansion` or `EnzoMethodGrackle`) should check for these parameters. In principle they should check both places, but this strategy will not scale well as we add more solvers in the future...
3. Relatedly some fluid-related parameters (like the choice of molecular weight or the temperature floor) were defined in the `ppm` solver section even though they were never used by the `ppm` solver. This again creates headaches if you need these parameters, but don't want to use the `ppm` solver.

## New Parameter File configuration
Since these fluid parameters are tracked in `EnzoPhysicsFluidProps`, we now specify them in the `Physics:fluid_props` section of the parameter file.  I'll provide some more details down below, but I would suggest you check out the examples and the new website documentation that is getting added by this PR. Since this `Physics:fluid_props` section is going to be present in just about every parameter file I made a conscious choice to make the format cleaner and more organized at the cost of having slightly more complex parsing code in `EnzoConfig`.

I took a lot of care to ensure that we maintained backward comparability with older versions of the configuration files. To facillitate this backwards compatibility I needed to introduce a minor "hack": Enzo-E always constructs an instance of `EnzoPhysicsFluidProps`, even when it is not explicitly specified in the parameter file. This leads to a quirk where the `Physics:fluid_props` parameters are honored even when `Physics:list` doesn't have an entry for `"fluid_props"` (which definitely contradicts the default behavior of the rest of the parameter file). If you think it's important, I can probably have Enzo-E abort with an error in this circumstance.

It would be helpful if we could drop some of this backwards-compatibility code in the near future since it definitely complicates things. 

## Configuring the Dual-Energy Formalism
This bears slightly more discussion. The Ppm and VL+CT solvers are compatible with different formulations of the dual energy formalism. As a result, we need to support 3 different configurations.
1. `"disabled"` - the dual-energy formalism is not in use
2. `"bryan95"` - this is the version of the dual-energy formalism that everybody is familiar with (it's what the Ppm solver uses). This is configured by 2 parameters: $\eta_1, \eta_2$.
5. `"modern"` - this is the formulation that the VL+CT solver uses. When I was implementing that solver @gregbryan and I decided to use a formulation more to that of Enzo's Runge-Kutta Solver (there is only a single criterion for synchronizing the internal and total energies). This is configured by a single parameter $\eta$. Note: the way that internal energy source terms are calculated is more similar to the `"bryan95"` approach than the that of Enzo's Runge-Kutta solver. At this time, there is a detailed description of this approach [here](https://enzo-e.readthedocs.io/en/latest/user/problem_method.html#dual-energy-formalism). 

I definitely welcome any suggestions for a better name for the `"modern"` configuration. The current name suggest that it might be *better* than the `"bryan95"` approach (even though that's not really true - they're just different). Long-term it might be nice to get rid of one of the dual-energy formulations, but that's probably a discussion for another time (As an aside, we can't really use the `bryan95` approach with the VL+CT solver at this time because it would significantly increase the required number of ghost zones...)

## Examples

I have included some examples to illustrate how the changes introduced in this PR will affect the configuration files (these examples are hidden and may not be visible if you're reading this over email)

<details> 
  <summary>VL+CT Example </summary>

   Configuration with the current version of the code:
   ```
    Field {
       gamma = 1.6666666666666667;
       # other parameters...
    }

    Method {
       list = ["mhd_vlct"];

       mhd_vlct {

          # quantity floors (these are used within the solver)
          density_floor = 1.e-6;
          pressure_floor = 1.e-6;

          # dual-energy formalism. This variation is parameterized by 1 variable
          dual_energy = true; # default is false
          dual_energy_eta = 0.001; # this is the default parameter

          # vlct-specific paramters:
          # ------------------------
          courant = 0.4;

          # the following are sensible values for a pure-hydro run
          mhd_choice = "no_bfield";
          riemann_solver = "hllc";
          half_dt_reconstruct_method = "nn";
          full_dt_reconstruct_method = "plm";
          theta_limiter = 1.5;
       }
    }
   ```

   An equivalent way of specifying the above parameters using the changes introduced by this PR (note, I have made sure we will continue to support the above approach for the near future).

   ```
    Physics {
       list = ["fluid_props"];

       fluid_props {
          dual_energy {
             type = "modern"; # to use the dual-energy formalism with VL+CT, need to use the
                              # "modern" formulation
             eta = [0.001]; # default value for dual_energy:type = "modern"
          }

          eos {
             # in the future we might allow configuration of other equations of
             # state, like an isothermal eos. 
             gamma = 1.6666666666666667;
          }

          floors {
             density  = 1e-6;
             pressure = 1e-6;
             # in the current version of the code, there isn't any way to specify the following parameter
             # (the parameter is used by some code outside of the VL+CT solver)
             temperature = 1e-6;
          }

          # in the current version of the code, there isn't any way to specify the following parameter
          # (the parameter is used by some code outside of the VL+CT solver)
          mol_weight = 0.6; # default value is 0.6
       }
    }

    Method {
       list = ["mhd_vlct"];

       mhd_vlct {
          courant = 0.4;

          # the following are sensible values for a pure-hydro run
          mhd_choice = "no_bfield";
          riemann_solver = "hllc";
          half_dt_reconstruct_method = "nn";
          full_dt_reconstruct_method = "plm";
          theta_limiter = 1.5;
       }
    }
   ```
</details>

<details> 
  <summary>PPM Example </summary>

   Current version of the code:
   ```
    Field {
       gamma = 1.6666666666666667;
       # other parameters...
    }

    Method {
       list = ["ppm"];

       ppm {

          # quantity floors. These are currently ignored within the solver. Any
          # time the solver places a floor, it uses a value of 1e-6 in code
          # units. However, these quantities are often used outside of
          # EnzoMethodPpm (e.g. see EnzoInitialIsolatedGalaxy)
          density_floor = 1.e-6;
          pressure_floor = 1.e-6;
          temperature_floor = 1e-6;

          # like the above parameters, EnzoMethodPpm never uses this. But,
          # other parameters do use this
          mol_weight = 0.6; # this is the default value

          # dual-energy formalism.
          dual_energy = true; # default is false
          dual_energy_eta_1 = 0.001; # this is the default value
          dual_energy_eta_2 = 0.1;   # this is the default value

          # ppm-specific paramters:
          # ------------------------
          courant = 0.8;

          # all of the following parameters are set to their default values
          diffusion = false;
          flattening = 3;
          minimum_pressure_support_parameter = 100;
          pressure_free = false;
          steepening = false;
          use_minimum_pressure_support = false;
       }
    }
   ```

   An equivalent way of specifying the above parameters using the changes introduced by this PR (note, I have made sure we will continue to support the above approach for the near future).

   ```
    Physics {
       list = ["fluid_props"];

       fluid_props {
          dual_energy {
             type = "bryan95"; # to use the dual-energy formalism with Ppm, need to use the
                              # "bryan95" formulation
             eta = [0.001, 0.1]; # default value for dual_energy:type = "bryan95"
          }

          eos {
             # in the future we might allow configuration of other equations of
             # state, like an isothermal eos
             gamma = 1.6666666666666667;
          }

          floors {
             density  = 1e-6;
             pressure = 1e-6;
          }

          mol_weight = 0.6; # default value is 0.6
       }
    }

    Method {
       list = ["ppm"];

       ppm {
          courant = 0.8;

          # all of the following parameters are set to their default values
          diffusion = false;
          flattening = 3;
          minimum_pressure_support_parameter = 100;
          pressure_free = false;
          steepening = false;
          use_minimum_pressure_support = false;
       }
    }
   ```
</details>

For completeness, I have also included a sample `fluid_props` configuration that doesn't use the dual-energy formalism, but is otherwise similar to the above configurations.

<details> 
  <summary>Sample without Dual Energy Formalism </summary>

   ```
    Physics {
       list = ["fluid_props"];

       fluid_props {
          dual_energy { type = "disabled"; }

          eos {
             # in the future we might allow configuration of other equations of
             # state, like an isothermal eos
             gamma = 1.6666666666666667;
          }

          floors {
             density  = 1e-6;
             pressure = 1e-6;
          }

          mol_weight = 0.6; # default value is 0.6
       }
    }
   ```
</details>

## Outstanding To Do List

The 2 outstanding items on this PR include:

- [x]  adding more detail to this message (explaining the changes in more detail and motivating why they are useful)
- [x] finish documenting the new approach for specifying fluid-property parameters.
- [x] there's a minor oversight related to configuring the dual-energy formalism in the code that needs to be addressed

## Future Plans

To get yt's derived fields to work appropriately, I'm going to need to tweak the frontend so that it knows it check the `"fluid_props"` section of the parameter files for details about the adiabatic index and whether the dual-energy formalism is in use (so that the front-end knows whether it should use the `"internal_energy"` field to compute some derived fields). This should all be very straight-forward to do. At this time, all tests depending on yt should continue to work without this tweak (as long as you aren't computing derived fields).

In a subsequent PR, it's my plan to integrate this better with some of the EOS machinery (basically, the `EnzoPhysicsFluidProps` object would store an instance of an EOS object).


mabruzzo:
@peeples This is now ready for review. There is one failing test that I need to address, but I'm confident that it will be a very minor fix (reviewer's shouldn't be worried about it at all)

mabruzzo:
I'll fix this last merge conflict shortly (I'm busy for the next 2 hours). Please don't let the conflict stop you from reviewing this PR

src/Enzo/enzo_EnzoConfig.cpp
```diff
@@ -2031,12 +1986,295 @@ void EnzoConfig::read_physics_(Parameters * p)
...
+    double pressure_floor = 0.0;
+    double temperature_floor = 0.0;
+    double metal_mass_frac_floor = 0.0;
+
+    auto get_ptr_to_floor_var = [&](const std::string name)
+      {
+        if (name == "density") { return &density_floor; }
+        if (name == "pressure") { return &pressure_floor; }
+        if (name == "temperature") { return &temperature_floor; }
+        if (name == "metalicity") { return &metal_mass_frac_floor; }
```
gregbryan, line 2140:
```suggestion
        if (name == "metallicity") { return &metal_mass_frac_floor; }
```

src/Enzo/enzo_EnzoMethodMHDVlct.cpp
```diff
@@ -36,20 +36,31 @@ static str_vec_t concat_str_vec_(const str_vec_t& vec1, const str_vec_t& vec2){
...
-				      double dual_energy_formalism_eta,
 				      bool store_fluxes_for_corrections)
   : Method()
 {
+  // check compatability with EnzoPhysicsFluidProps
+  EnzoPhysicsFluidProps* fluid_props = enzo::fluid_props();
+  const EnzoDualEnergyConfig& de_config = fluid_props->dual_energy_config();
+  ASSERT("EnzoMethodMHDVlct::EnzoMethodMHDVlct",
+         "selected formulation of dual energy formalism is incompatible",
+         de_config.is_disabled() | de_config.modern_formulation());
```
gregbryan, line 49: Do you really want to use the bitwise | operator here (why not ||)? Feel free to ignore if I am missing something.

src/Enzo/enzo_EnzoMethodPpm.cpp
```diff
@@ -40,6 +40,13 @@ EnzoMethodPpm::EnzoMethodPpm (bool store_fluxes_for_corrections)
     store_fluxes_for_corrections_(store_fluxes_for_corrections)
 {
 
+  // check compatability with EnzoPhysicsFluidProps
+  EnzoPhysicsFluidProps* fluid_props = enzo::fluid_props();
+  const EnzoDualEnergyConfig& de_config = fluid_props->dual_energy_config();
+  ASSERT("EnzoMethodPpm::EnzoMethodPpm",
+         "selected formulation of dual energy formalism is incompatible",
+         de_config.is_disabled() | de_config.bryan95_formulation());
```
gregbryan, line 48: same comment as earlier about or operator

gregbryan (approved):
Looks good to me -- I think there may be a few very minor issues that need to be looked at (see comments) but once those are addressed, this is ready to go.

src/Enzo/enzo_EnzoMethodMHDVlct.cpp
mabruzzo: I'll change this since it's confusing here. 

This is a weird habit I've picked up (I've started to use the bitwise `&`  and bitwise `|` operators instead of `&&` and `||` unless I really need the short-circuiting functionality. But, it would be better to use the logical operators by-default)

src/Enzo/enzo_EnzoConfig.cpp
mabruzzo: Thanks for catching this. I'll be sure to fix this!

doc/source/param/physics.incl
```diff
@@ -61,3 +74,170 @@ cosmology
...
+ideal gas. But, in the future this will support alternative 
+equations of state (like an isothermal equation of state).
+
+.. note::
+
+   `PR #173 <https://github.com/enzo-project/enzo-e/pull/173>`_
+   introduces a discussion about how the adiabatic index is treated
+   when the ``"grackle"`` method is in use and the :p:`Method` :
+   :p:`grackle` : :p:`primordial_chemistry` parameter exceeds 1.  This
+   should link to that discussion after the other PR is merged.
```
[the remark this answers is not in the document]
mabruzzo: Thanks for catching this. I'll be sure to fix this!

PR #187, opened 2022-05-30
WIP: Cosmology support for VL+CT

mabruzzo:
This is a WIP PR for making the VL+CT solver support cosmology, when run without magnetic fields. This builds off of PR #186.  I'm pretty confident in the correctness of my implementation.

To try to test it, I tried to quickly port the Zeldovich Pancake test. Unfortunately, this took a lot longer than I expected and I ran into some issues. I'm not sure when exactly I'll have time to get back to this, so I wanted to post my progress so far.

The following shows what happens over the first 8 cycles of the simulation when initializing an x-axis aligned Zeldovich Pancake:
image
To reproduce this plot, use this input file at `input/Cosmology/ZeldovichPancake/vlct-HD-x-aligned.in` and [this python script](https://pastebin.com/ew23ZLsQ).

In the final snapshot, you can see that we start to enforce the density floor near the edges of the domain. The bottom row shows the relative difference between the total energy and the kinetic energy. In the second to-last snapshot this goes to a negative number on the domain edge which suggests we're encountering the problem that the dual-energy formalism is supposed to address. For reference, this simulation doesn't currently use the dual-energy formalism (we need a minor change introduced in PR #173 for `EnzoMethodComovingExpansion` to properly support the VLCT solver with the dual-energy formalism - but I can manually port that to this PR as well).

I suspect that I made a transcription error in porting the Zeldovich Pancale initializer. I was definitely not very careful (especially about units). However, it would be helpful if someone knowledgable about the test in Enzo can:
- comment if the dual-energy formalism is strictly necessary for this test
- potentially provide me with some snapshots of the Enzo results (on a uniform grid) that I can eventually use for comparisons

If people really want this feature sooner, we could shelve this test for now and come up with something simpler...


gregbryan:
This doesn't look crazy in terms of shape, but units are worrying -- the package should have a cosmic mean density, and at z=20 that is much higher than what is given (assuming CGS units). The dual energy formalism was developed and tested on *this* problem, so it's pretty useful (as shown in the bottom panel -- note that just because this is positive doesn't mean it's right). I couldn't quickly dig up Enzo plots but happy to do so once you get back to this. It's just a matter of running the enzo test problem and plotting the same outputs.

mabruzzo:
There's a dual-energy related issue. This is not going to be done for the v1.0 release

PR #192, opened 2022-06-03
Address minor oversight in the ported tests

mabruzzo:
In the process of porting the C++ unit tests and the tests that compare the final simulation time or final cycle to the actual simulation time/cycle at the end of an Enzo-E simulations, there was a minor oversight  that reduced the functionality of these tests. 

The new system implicitly assumes that the exit code of these C++ tests always indicated whether the test passed or not. However, that is not presently the case (as long as the program gracefully exits, it reports an exit code of 0). As a consequence, a test could fail (e.g. the final simulation time could be different from the expected final time), but CTest would still report the test as passing. As a concrete example, one can do the following:
- modify line 79 of `input/PPM/ppm.incl` so that you replace `time_final  = [1.04069762624371];` with something like `time_final = [10.0];`.
- then when you execute `ctest --verbose -R PPM-1` from the build directory, `ctest` reports that the test failed, even though the output clearly states the final time of the simulation is incorrect.

My first instinct to solve this was to pipeline running the simulation, dumping stdout to a log file, and then call a script that parse the logfile and exits with the appropriate exit code. Unfortunately, the way that ctest works requires that a test is entirely encapsulated by a single command (that takes optional potential arguments).

To that end, I have introduced a python script called `tools/run_cpp_test.py` to do all 3 steps for us at once. The script wraps and executes an arbitrary command (much like a profiling tool like `time` or `perf-stat`) and then parses the standard output stream to check for indications that the test may have failed (I have ported this logic from the `build.sh` script from just before we transitioned to cmake).

As a concrete example, one could execute the following command from the repository's root directory:
```
python3 tools/run_cpp_test.py --output-dump sim_output.log path/to/enzo-e input/PPM/method_ppm-1.in
```
Note that the optional `--output-dump` argument specifies where the script writes a copy of stdout (It still continues to write a copy of stdout to stdout). This is what the script parses to check if the program succeeded. If you don't specify this, the program will write a copy of stdout to a temporary file that will be deleted once the program ends.

There are 2 other alternative approaches for addressing this problem:
1. We continue to use the wrapper script approach, but rewrite it in a different language (like bash or c++)
2. Modify the unit testing machinery to always exit with a non-zero exit code if it detects a testing error. Likewise, we would need to modify Cello/Enzo-E to exit with a non-zero exit code if the testing machinery detects an unexpected completion time/completion cycle.

I actually really like the 2nd approach as a longer term solution. But, that will take some additional work and this PR at least offers a near-term solution.

### DIFF
--- a/tools/run_cpp_test.py
+++ b/tools/run_cpp_test.py
@@ -26,7 +26,8 @@ parser.add_argument(
     help = "the C++ binary that is to be executed"
 )
 parser.add_argument(
-    "args", metavar = "ARGS", action = "store", nargs = '*', default = [],
+    "args_for_command", metavar = "ARGS", action = "store", nargs = '*',
+    default = [],
     help = "the arguments to the C++ binary that are to be executed"
 )
 
@@ -102,7 +103,8 @@ if __name__ == '__main__':
     else:
         dump_path = args.output_dump
 
-    success = execute_command(command = args.command, args = args.args,
+    success = execute_command(command = args.command,
+                              args = args.args_for_command,
                               output_dump = dump_path)
     if not success:
         out = 1
--- a/tools/run_cpp_test.py
+++ b/tools/run_cpp_test.py
@@ -1,0 +1,118 @@
+import argparse
+import subprocess
+import sys
+import tempfile
+
+_description = '''\
+Runs a test that is directly encoded in a C++ binary (the binary is specified
+by COMMAND and optional arguments are specified by ARGS), that makes use of the
+Cello testing machinery. After the test completes, this program determines
+whether it was succesful and returns an appropriate exit code (an exit code of
+0 indicates the test was entirely successful).
+'''
+
+
+parser = argparse.ArgumentParser(description = _description)
+parser.add_argument(
+    "--output-dump", action = "store", default = None,
+    help = ("Specifies the path where a copy of the standard output stream "
+            "from the execution of the program should optionally be dumped "
+            "(the data is still written to the standard output stream). The "
+            "contents of this file may be used to determine the outcome of "
+            "the tests.")
+)
+parser.add_argument(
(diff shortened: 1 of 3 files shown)


mabruzzo:
@peeples this is ready for review. 

As an aside, I think this is a very high-priority PR because it fixes a bunch of tests that were unintentionally broken in the transition to CMake. Without the changes from this PR, a bunch of tests can fail and we would have no way of knowing it from the continuous integration checks (the testing framework will misreport the test results).

gregbryan (approved):
Thanks Matthew -- this looks like a sensible solution. Agreed that a better long-term solution would be to have the test itself return its status (so we don't have to resort to searching the output for clues), but for now, I think this solves the issue.

tools/run_cpp_test.py
```diff
@@ -0,0 +1,118 @@
...
+
+if __name__ == '__main__':
+    args = parser.parse_args()
+
+    if args.output_dump is None:
+        dump_path = tempfile.mktemp() # path for a temporary file
+    else:
+        dump_path = args.output_dump
+
+    success = execute_command(command = args.command, args = args.args,
```
stefanarridge, line 105: Personally I think `args = args.args` is a little confusing. Maybe call the arguments which are passed to the C++ binary something like `cpp_binary_args`. Then again, this would make things a bit more verbose. I'll let you decide.

stefanarridge (approved):
Looks good to me - only have one small suggestion.

I think your idea for how to improve the tests going forwards is a good one, should probably discuss at the workshop.

tools/run_cpp_test.py
mabruzzo: Good point. I renamed it `args_for_command`

mabruzzo:
Once all these tests pass, this can be merged

PR #196, opened 2022-06-09
Bugfix for "merge_sinks" tests 

stefanarridge:
Fixed bug in `"merge_sinks"` tests: input parameter file now specifies that sink particles are in the `"is_gravitating"` group, which ensures that the `"pm_update"` method updates the particle positions, which means that the particles move towards each other and merge.


peeples:
@fearmayo @mabruzzo this should be quick

mabruzzo (approved):

jwise77 (approved):

Issue #197, opened 2022-06-09
Problems with mixed precision between fields and particles


jwise77:
When I was debugging the particles in the IsolatedGalaxy test, I discovered that it would not run when the user-specified particle attribute precision did not match the precision of `enzo_float`. These crashes occur when the particle attribute is cast into the `enzo_float` type in a call like
```c++
  px    = (enzo_float *) particle.attribute_array(it, ia_x, ib);
```
that would result in garbled data when `type(enzo_float) != type(particle.attribute_type)`.

In the IsolatedGalaxy initializer, I worked around this issue by writing the following routine
```c++
template <class T>
void EnzoInitialIsolatedGalaxy::assign_attribute_value
(Particle * particle, T value, int it, int ia, int gindex)
{
  if (ia < 0) { // undefined
    return;
  }

  int ib = 0;
  int ipp = 0;

  particle->index(gindex, &ib, &ipp);

  // Get pointer of the attribute array
  char * attr = particle->attribute_array(it,ia,ib);

  // Get type of attribute array
  int type = particle->attribute_type(it,ia);

  if (cello::type_is_float(type)) {
    if (type == type_single) {
      float * attr_f = (float*) attr;
      attr_f[ipp] = value;
    } else if (type == type_double) {
      double * attr_d = (double*) attr;
      attr_d[ipp] = value;
    } else if (type == type_quadruple) {
      long double * attr_q = (long double*) attr;
      attr_q[ipp] = value;
    } else {
      ERROR1("EnzoInitialIsolatedGalaxy.C // assign_attribute_value():",
      "Unknown particle attribute float type %d", type);
    }
  } else if (cello::type_is_int(type)) {
    if (type == type_int8) {
      int8_t * attr_8 = (int8_t *) attr;
      attr_8[ipp] = value;
    } else if (type == type_int16) {
      int16_t * attr_16 = (int16_t *) attr;
      attr_16[ipp] = value;
    } else if (type == type_int32) {
      int32_t * attr_32 = (int32_t *) attr;
      attr_32[ipp] = value;
    } else if (type == type_int64) {
      int64_t * attr_64 = (int64_t *) attr;
      attr_64[ipp] = value;
    } else {
        ERROR1("EnzoInitialIsolatedGalaxy.C // assign_attribute_value():",
        "Unknown particle attribute int type %d", type);
    }
  }
```

[...]

PR #202, opened 2022-06-13
Changing single/double particle precision in parameter files to default

jwise77:
As discussed in #197, a precision mismatch can happen between the particle attributes and fields. A short-term solution is to set the attribute precision to "default" instead of single or double in the offending parameter files, which this PR addresses.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,7 @@ charmrun
 config.log
 errors.org
 warnings.org
+doc/dox-xml
 
 # ignore directories used for building branches
 build-*/Cello/*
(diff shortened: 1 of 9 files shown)


mabruzzo (approved):

stefanarridge (approved):

jwise77:
The tests in `Particles/` are failing because something is hard-coded that assumes single precision. I'm looking into it now.

jwise77:
All is good and the tests pass now after reverting back to single precision for the tracer particles.

Issue #207, opened 2022-06-13
Test of Collapse methods fails


buketbenek:
Testing of collapse method gives MPI_Abort(). @stefanarridge recommended updating Line 58 of pm2.incl to
constants = ["mass", "double", 1.2207031e29] for Particle. It fails at Enzo-E final cycle with this setting.

stefanarridge:
I suggested using the value of 1.2207031e29 since that is 2e33 multiplied by the root level cell volume (actually area in this case), which is 1/128^2 for this problem.

It seems that this input file is not currently being used by the test suite, since this was not picked up when PR #89 was merged into `main`.

buketbenek:
I just checked again and it fails at the final cycle with mass and 2e33 settings.

stefanarridge:
Ah okay... that's not good...

stefanarridge:
Okay, try and replace the value with `3.2768e+37`, which is 2e33 *128^2.

buketbenek:
With 3.2768e+37, it gives:
0 00003.01  @@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
0 00003.01 ERROR control_refresh.cpp:1387
0 00003.01 ERROR Block::particle_scatter_neighbors_
0 00003.01 ERROR particle indices (ix,iy,iz) = (3,10,0) out of bounds
0 00003.01  @@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@

PR #214, opened 2022-06-14
Fixing particle mass in test_cosmo.incl

WillHicks96:
Very small fix to test_cosmo.incl so that it runs -- initializing DM mass as "mass" (not "density"). These files didn't get updated after PR #89 merged

### DIFF
--- a/input/test_cosmo.incl
+++ b/input/test_cosmo.incl
@@ -283,7 +283,7 @@ Adapt {
  Particle {
      dark {
          attributes = [ "x", "default", "y", "default", "z", "default", "vx", "default", "vy", "default", "vz", "default", "ax", "default", "ay", "default", "az", "default", "is_local", "default"];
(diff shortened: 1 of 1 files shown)


stefanarridge:
This seems to be related to Issue #207 (which mentions a different file).

TA few things to point out:
- These files must not be used by any of the automated tests, which is why this wasn't picked up when reviewing PR #89. Are these files needed in that case, or are there tests which are currently missing?
- The file which includes this file doesn't have a `Testing` parameter group, so I'm not sure how to check if this runs correctly (does the output look sensible?)
- I mention in the comments of Issue #207 that I suggested using a different value for the `"mass"` / `"density"`, but the test still failed, which is a bit worrying. I am going to suggest running it again with the value of `2e33`. That could work.

jwise77 (approved):
I'm approving, but I think @stefanarridge makes some excellent points that are now linked in #207

PR #221, opened 2022-06-15
[BUGFIX] Isolated galaxy velocity calculation

jwise77:
In the original version of the initializer, the halo mass can be derived from the central DM density. This is triggered by giving a negative halo mass and a non-zero DM density. However, there was one calculation in the initializer that wasn't checking whether the halo mass was negative, resulting in a NaN after a sqrt. This PR fixes this issue.

### DIFF
--- a/input/IsolatedGalaxy/WLM_small.in
+++ b/input/IsolatedGalaxy/WLM_small.in
@@ -133,7 +133,6 @@
       #    also can have background stellar bulge and / or stellar disk potential
       background_acceleration {
           flavor = "GalaxyModel";
-          DM_density     = -1.0   ;  # specify this OR mass - not both - mass_radius must be virial if this is used
           DM_mass        = 1.0E10;   # mass interior to below radius (in Msun) -  (this particular value is virial mass)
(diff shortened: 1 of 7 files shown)


gregbryan (approved):

stefanarridge (changes requested):
I know this has already had two approvals, but I am a bit confused about something.

You say that "In the original version of the initializer, the halo mass can be derived from the central DM density". But it looks like what was actually happening was that the mass was being set by the user in the parameter file. Now, with your changes, it is the "DM density" that is set by the user, and the "DM_mass" value is overwritten. If the "DM mass" parameter is no longer used, I would suggest removing it.

Also, I would suggest changing `rho_dm`, `rcore`, `rvir`,`Mvir` to `rho_dm_cgs`, `rcore_cgs`, `rvir_cgs`, `Mvir_cgs`.

Edit: Okay, I have actually read the comment in the code now and I see that `Mvir` is only overwritten if it is negative.

I'd still recommend removing the `"DM mass"` parameter, and forcing the user to use a positive value for "DM density", which can be done with an `ASSERT` call.

stefanarridge (approved):

PR #126, opened 2021-08-25
Add report and modifications for gravity solver optimization

KillerKaninchen:
Pull request to add the results of the gravity solver optimisations performed by a team from Maynooth University and ICHEC.

It consists of a document in the general documentation outlining the results of the performance analysis and small changes to the code base to fix the `min_face_rank` in the Jacobi and Multigrid solver to 2.

PR #229, opened 2022-06-16
[BUGFIX] Removed line of code which caused compilation to fail

stefanarridge:
Simple bugfix, which came down to removing a call to a function that no longer exists, which somehow came about when PR #126 was merged into `main`.

I presume it is safe to remove, since none of the other `Solver` classes call the function in their constructor methods, but perhaps @jobordner should confirm.


fearmayo (approved):
Yes I believe that line got merged in accidently. I'm not sure how - I removed it in the latest commit. Something funky must have happened when merging.

gregbryan (approved):

PR #234, opened 2022-06-20
Cosmology units

stefanarridge:
Updated EnzoPhysicsCosmology to use values of physical constants from enzo_constants to compute units, as well as adding descriptive constants explaining the choice of units.

I am still somewhat confused about the time units, specifically the factors of `(1+initial_redshift_)`. My reasoning is that it has to do with how many factors of `a` appear in the gravitational potential (see Equations 7, 17, and 18 of the Enzo paper: https://arxiv.org/pdf/1307.2265.pdf), but Equation 7 does not seem to match up with the code, specifically lines 178-179 of `src/Enzo/enzo_EnzoMethodGravity.cpp` (there is no division by `a` from what I can tell).

There is also a slight discrepancy with the old and new versions of the computation of the density unit, which I suspect has something to do with using different values for G.


src/Enzo/enzo_EnzoPhysicsCosmology.hpp
```diff
@@ -180,48 +180,110 @@ class EnzoPhysicsCosmology : public Physics {
...
   }
 
-  double velocity_units() const
-  {
-    return 1.22475e7*comoving_box_size_*sqrt(omega_matter_now_)*
-                      sqrt(1.0 + initial_redshift_);
-  }
+  // Return velocity unit in cm/s
+  // Equation 17 of Greg L. Bryan et al 2014 ApJS 211 19, shows that the dimensions of velocity
+  // are `a` times `comoving length` over `time`.
```
mabruzzo, line 283:
```suggestion
  /// Return velocity unit in cm/s
  /// Equation 17 of Greg L. Bryan et al 2014 ApJS 211 19, shows that the dimensions of velocity
  /// are `a` times `comoving length` over `time`.
```
This just fixes doxygen docstring formatting

src/Enzo/enzo_constants.hpp
```diff
@@ -45,6 +45,8 @@ namespace enzo_constants {
   const double clight = 29979245800.0;
 
   // Gravitational constant in CGS
+  // Note: this value is ovewritten by Method:gravity:grav_constant in non-cosmological
+  // simulations.
```
mabruzzo, line 49:
```suggestion
  // Note: in non-cosmological simulations, the Gravity solver uses the value specified by 
  // Method:gravity:grav_constant
```
Minor suggestion (to avoid the word `overwrite`), feel free to ignore this.

doc/source/user/problem_units.rst
```diff
@@ -0,0 +1,27 @@
...
+the mean comoving matter density is given by :math:`\frac{3 H_0^2 \Omega_m}{8 \pi G}`.
+
+The time unit is defined so that :math:`\frac{3}{2} H_0^2 \Omega_m (1+z_i)^3 = 1`, where
+:math:`z_i` is the initial redshift of the simulation.
+
+For cosmological simulations, the value set for ``Method:gravity:grav_const`` is ignored.
+
+
+
+
```
mabruzzo, line 27:
```suggestion
In all simulations, the ``"temperature"`` field always has units of Kelvin.
```
Minor suggestions

mabruzzo (changes requested):
I didn't look super-carefully at the math (I just kind of assumed that the calculations were correct since the tests passed).

There are 3 main formatting changes I would like to see related to doxygen docstring formatting before. I also main a 2 very minor suggestions, but feel free to ignore those

src/Enzo/enzo_EnzoPhysicsCosmology.hpp
```diff
@@ -180,48 +180,110 @@ class EnzoPhysicsCosmology : public Physics {
...
-  /// units of Kelvin to units of specific internal energy
-  ///
-  /// The original Enzo refered to this quantity as "temperature units", but in
-  /// Enzo-E we primarily track temperature in units of Kelvin
-  double kelvin_per_energy_units() const
+  double velocity_units() const
   {
-    return 1.81723e6*pow(comoving_box_size_,2.0)*omega_matter_now_*
-                      (1.0 + initial_redshift_);
+    return length_units() / time_units() * (1 + initial_redshift_) / (1 + current_redshift_);
```
gregbryan, line 286: I think this definition is fine and is certainly clearer. The one potential downside is that it is not immediately obvious from inspection that velocity units are independent of time. I think it would be worth adding this as a comment. The other potential issue is that the exact value returned by this routine may change by very small amounts of time due to roundoff error (the 1+z terms will not cancel exactly) -- I don't think this is a problem but just wanted to point it out.

gregbryan (changes requested):
Thanks -- a few suggestions in the code.

gregbryan:
The time units are defined such that the free-fall time at z=z_i is 1.0. That gives the Poisson equation its simple form. Regarding the Poisson equation (eq. 8 in the Enzo method paper I think) -- I believe that the missing factor of 1/a you mention is added on line 266 of `enzo_EnzoMethodGravity.cpp`.

The slight inconsistency of the density unit comes from the carelessness of past generations...

stefanarridge:
>I believe that the missing factor of 1/a you mention is added on line 266 of `enzo_EnzoMethodGravity.cpp`.

I was under the impression that that factor of `1/a` on line 266 corresponded to equation 18, since it is part of the `compute_accelerations` function.

src/Enzo/enzo_EnzoPhysicsCosmology.hpp
```diff
@@ -180,48 +180,121 @@ class EnzoPhysicsCosmology : public Physics {
...
+
+    return sqrt(2.0/3.0) / (enzo_constants::H0_over_h * hubble_constant_now_ *
+			    sqrt(omega_matter_now_ *
+				 (1 + initial_redshift_) *
+				 (1 + initial_redshift_) *
+				 (1 + initial_redshift_)));
   }
 
+  /// Return velocity unit in cm/s
+
```
mabruzzo, line 282:
```suggestion
```

mabruzzo (approved):
Thanks for making those changes. I have one last minor formatting suggestion - but otherwise, it looks great to me!

gregbryan:
> > I believe that the missing factor of 1/a you mention is added on line 266 of `enzo_EnzoMethodGravity.cpp`.
> 
> I was under the impression that that factor of `1/a` on line 266 corresponded to equation 18, since it is part of the `compute_accelerations` function.

And I thought that the factor of 1/a from eq. 18 was implemented in lines 115-117 of enzo_EnzoComputeAcceleration.cpp (but I admit I haven’t gone over this part of the code carefully).

stefanarridge:
> And I thought that the factor of 1/a from eq. 18 was implemented in lines 115-117 of enzo_EnzoComputeAcceleration.cpp (but I admit I haven’t gone over this part of the code carefully).

Yes, I believe you're right.

gregbryan (approved):
Thanks for making those changes and clarifying the units -- many future users will appreciate that!

PR #248, opened 2022-07-01
Updating build-system to re-add unit tests

mabruzzo:
The main changes in this PR include: 
- I systematically updated the dependencies for each Cello component
- I introduced a handful of extra internal Cello targets that define machinery specially geared towards certain unit tests (these targets are compiled from `main.cpp`, `main_mesh.cpp`, and `main_simultation.cpp`) 
- I also slightly refactored the CMakeLists.txt script in the Enzo layer

As a result, I've managed to resolve all of the linking problems. We've gone from just being able to compile 6 unit tests (of a total of 38 unit tests) to now being able to compile 37 unit tests. I also fixed a broken test and a very minor bug that crept into `EnzoPhysicsCosmology` (that probably only manifests in the unit test).

Two tests require further attention (which I have documented in issues #254 and #255) I've provided a summary below:
1. The `"test_Index.cpp"` test will not compile. It appears that the API of the `Index` class got altered, but the test was not updated accordingly (since there were problems with building the tests). It's not completely obvious to me how this should be fixed. To see the exact compilation problems, you need to uncomment the following line at the end of `src/Cello/CMakeLists.txt`:
    ```cmake
    #addUnitTestBinary(test_index "test_Index.cpp" mesh tester_mesh)
    ```
    and then you should execute `make test_index` (or `ninja test_index`) from your build directory.

2.  The `test_adapt` test currently fails (even though it compiles). To build this test, you can call `make test_adapt` (or `ninja test_adapt`) from the build directory. Then, you can run the test by calling `./bin/test_adapt` from the build directory. This could potentially be indicative of a problem. However, I've disabled this from the automated testing framework right now (since the failure of this test test should not block the rest of the changes).


peeples:
@jwise77 @gregbryan Just making sure you saw this one

gregbryan (approved):
Looks good to me -- great to have those unit tests back in (although I am very far from being a CMAKE expert!)

jwise77 (approved):
All looks good to me. Thanks for this update!

PR #253, opened 2022-07-07
Fix bug from PR#239 in EnzoMethodPmDeposit

mabruzzo:
PR #239 introduced a minor bug in which array used for storing the output of dep_grid_cic was not large enough. This caused dep_grid_cic to write values to invalid locations in memory. This bug manifested in weird ways when using compiling with intel compilers (under special conditions)

In addition to fixing that bug, some variable names were updated so that their significance is more obvious


jobordner (approved):

gregbryan (approved):
This is likely due to the fact that CIC deposition can end up with mass off the edge of the active grid.

tumlinson (approved):
approved based on success of modest cosmo runs that had previously failed.

Issue #124, opened 2021-06-24
PPM Crashing in SMP Mode


WillHicks96:
I am getting segfaults in problems that use PPM whenever I run in SMP mode with  >1 PE per logical node. The segfault happens specifically in `flux_twoshock`.

With `input/PPM/method_ppm-8.in`, for example, running with the following command gives me [this output](https://github.com/enzo-project/enzo-e/files/6712576/ppm_smp_error.txt):

> `charmrun +p4 bin/enzo-e +ppn 2 input/PPM/method_ppm-8.in`

If I run with `+ppn 1` instead, the test runs through without any errors.

jobordner:
Have you tried with flux correction turned off?  Also `Initial : value` initialization is known to be flakey in SMP mode.

WillHicks96:
@jobordner Yes, I'm getting the same error even with flux correction turned off for the `method_ppm-8.in` test. 

I'm getting this behavior with `test_cosmo-dd-fc0.in` as well, which doesn't use the `value` parameter for initialization.

mabruzzo:
I might be stating the obvious here, but I'd be a little surprised if this is actually an issue with PPM. 

Try out [this file](https://pastebin.com/QqpyfFcP). This is basically the simplest possible test without using the `"value"` initializer (I'm fairly confident that the `"inclined_wave"` initializer is thread-safe).

WillHicks96:
@mabruzzo Still getting the same thing! Here's the output from your test: [inclined_wave_smp_error.txt](https://github.com/enzo-project/enzo-e/files/6713059/ppm_error_inclined_wave.txt)

mabruzzo:
Thanks for running that. I stand corrected: the issue definitely seems to be related to the PPM Solver. I seem to recall that it worked at some point. I wonder if it broke with PR #24 or #48 or if it was something else.

I was able to reproduce the error locally. I tried running [another version of the same simulation](https://pastebin.com/4Shm3n7n) where I replaced the PPM Solver with the "hydro-only" version of VL+CT. I haven't verified that the results are what we would expect, but it definitely doesn't crash. I plan to come back and look at this more this afternoon.

Note: you might not be able to run the newer version of the test unless your branch has some recent commits from master. If you're interested, you might be able to see that the following command works: `charmrun ++ppn 2 bin/enzo-e ./input/vlct/HD_linear_wave/method_vlct_soundN32.in`

[...]

PR #256, opened 2022-07-15
[Bugfix] CMake checks whether charm++ was compiled with SMP mode in multiple places.

mabruzzo:
Newer versions of charm++ (starting at least in version 7.0) appear to define the macro specifying SMP usage in a different header file than the one we checked.

Prior to this commit, CMake would always effectively assume that these newer versions of charm++ were built without SMP mode

### DIFF
--- a/cmake/FindCharm.cmake
+++ b/cmake/FindCharm.cmake
@@ -185,6 +185,19 @@ if(CHARM_COMPILER)
     include(CheckSymbolExists)
     CHECK_SYMBOL_EXISTS(CMK_SMP "${CHARM_INCLUDE_DIR}/conv-mach-opt.h"
                         CHARM_SMP)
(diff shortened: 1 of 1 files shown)


mabruzzo:
This has no bearing on whether this PR should be merged but, I did want to mention that bugs remain while running the PPM solver in SMP-mode (documented in Issue #124).
- One of the times that I ran `ctest`, the `PPM-8` test failed. However, I can't reliably reproduce this error. This might suggest a race-condition.
- The `AmrPpm` test reliably fails

I also noticed that the ``bb_test`` times out. I'll document this in a new issue.

All of these issues were documented using a gcc build.

jwise77 (approved):

gregbryan (approved):

PR #257, opened 2022-07-15
Fix CircleCi so that it runs VL+CT tests again

mabruzzo:
When PR #244 refactored the configuration file to break compilation and executing the ctest framework into separate build-steps, this was accidentally broken.

It's still my intention to port the VL+CT tests to the answer testing framework, but I don't have time right now. This will ensure that the tests continue running in the short-term.


jwise77 (approved):

brittonsmith (approved):

jwise77:
It seems like `USE_DOUBLE` isn't equal to either OFF or ON. It's throwing the error message
```
ERROR while checking precision of enzo_float
```

PR #264, opened 2022-07-28
Fix leaf init

jobordner:
Update Block `is_leaf` initialization to be true iff `(level >= 0)` instead of always true, otherwise `is_leaf` is true for negative-level blocks. Previously, there was still a check for negative levels in `Block::is_leaf()`, so most code still ran correctly; but the incorrect true values for negative level blocks were still written to checkpoint files, which could confuse analysis code.

Bonus bug fix: test_Memory was failing due to excessive optimization, so some printf's were added to the test code inhibit the optimization, so test_memory now passes.


gregbryan (approved):

mabruzzo (approved):
Looks good to me!

PR #265, opened 2022-08-16
Updated the developer's guide for writing `Method`s

mabruzzo:
These changes address some questions that I recently answered (I was reminded that while some of this stuff may seem obvious after you're familiar with the codebase, things can seem overwhelming at first).

I'm definitely open to feedback. This is not very polished, but seemed like a decent start.


jwise77 (approved):
Thanks for adding this! It all reads well to me.

gregbryan (approved):
Very useful!

PR #266, opened 2022-08-23
Enforcing density floor in EnzoMethodPpm

WillHicks96:
This is a small PR that adds the ability to set a density floor in EnzoMethodPpm, since the method doesn't currently check `enzo_fluid_floor_config.density()` anywhere.
 
I've also made the small change of passing density and pressure floors into `ppm_de` as parameters. In the main branch, these are currently hard-coded to be arbitrarily small numbers in `ppm_de` (for pressure_floor) and `<x/y/z>euler_sweep` (for density_floor).

### DIFF
--- a/src/Enzo/enzo_SolveHydroEquations.cpp
+++ b/src/Enzo/enzo_SolveHydroEquations.cpp
@@ -135,6 +135,24 @@ int EnzoBlock::SolveHydroEquations
 	    "Grid::SetMinimumSupport() returned ENZO_FAIL");
     }
   }
+
+  /* Set density floor. */
+  const EnzoFluidFloorConfig& fluid_floor_config
+               = fluid_props->fluid_floor_config();
+  enzo_float density_floor = fluid_floor_config.has_density_floor() ?
+                 fluid_floor_config.density() : 0.0;
+
+  if (density_floor > 0.0) {
+    for (int i=0; i<mx*my*mz; i++) {
+       if (density[i] < density_floor) {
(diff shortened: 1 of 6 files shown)


src/Enzo/enzo_SolveHydroEquations.cpp
```diff
@@ -135,6 +135,24 @@ int EnzoBlock::SolveHydroEquations
...
+
+  /* Set density floor. */
+  const EnzoFluidFloorConfig& fluid_floor_config
+               = fluid_props->fluid_floor_config();
+  enzo_float density_floor = fluid_floor_config.has_density_floor() ?
+                 fluid_floor_config.density() : 0.0;
+
+  if (density_floor > 0.0) {
+    for (int i=0; i<mx*my*mz; i++) {
+       if (density[i] < density_floor) {
```
gregbryan, line 147: Maybe use std::max here? (I think performance probably doesn't matter because of optimization but it just seems clearer).

gregbryan (approved):
Thanks for catching that! I just looked at ppm_de.F and saw it had the same density floor as enzo and assumed it was actually connected to the corresponding parameter...  One very minor suggestion (but feel free to ignore).

src/Enzo/enzo_SolveHydroEquations.cpp
WillHicks96: Change made!

mabruzzo (approved):
Looks great to me!

PR #267, opened 2022-08-25
Port Grackle Tests to Answer Testing and add Cosmo Answer Tests

mabruzzo:
This PR does 4 things related to the answer testing framework:

1. I made a few minor tweaks to our usage of f-strings to make the framework compatible with older versions of python
2. it adds machinery for supporting answer tests that need Grackle. 
    - My solution is not necessarily the most elegant, but I think it's **very** easy to use. Adding tests that use Grackle has always been fairly painful (since we don't store datafiles in this repository), but now it should be relatively easy to do.
    - I've updated the docs accordingly to explain how to use this machinery
3. I've ported the Grackle tests from the ctest framework to the answer testing framework (to employ this new machinery).
4. I've introduced a new cosmology test to the answer testing framework.
    - This test derives from `test_cosmo-dd-multispecies.in` and covers areas of the code not touched by other tests. It should let us avoid a class of bugs that seem to repeatedly arise whenever we add a new major feature
    - ~~Currently, this test just compares the final simulation time, which is better than the alternative~~
    - ~~There seems to currently be a bug in the yt-frontend with calling `ad.quantities.weighted_standard_deviation` for an AMR simulation (the error complains about the weight field and field having different shapes). I'll make an issue on the yt GitHub page documenting this problem.~~
    - **EDIT:** The test compares some summary statistics for various fields (the issues I encountered before were related to computing a standard deviation of a particle field with a fluid-weight-field)


gregbryan (approved):
Looks like a very useful addition.

brittonsmith (approved):

brittonsmith:
This looks great to me. Let's merge!

mabruzzo:
Great! I'm just going to add a the ``gold-standard-tag-002`` tag, and update CircleCi to use that when generating CircleCi results (with the way that the answer testing is set up these new tests won't be used until I do that). You can see that since the last time we ran CI the new tests were skipped (as an aside, it's unclear why the tests get skipped rather than fail... - they fail when I do this locally). 

Assuming that all tests pass after I do this, then I'll hit merge.

In the future, we may want to come up with a more stream-lined process for doing this... We generally want people to make sure that the existing tests all pass before they make this sort of update (i.e. to ensure that they haven't broken an existing test) and we don't generally want people editing their PR after it has been reviewed...

PR #269, opened 2022-11-17
VL+CT test fix

mabruzzo:
This fixes some failing tests of the VL+CT integrator, that arose because of a recent change in `yt` that affected the python script for computing the L1 Error Norm. I also did some light refactoring.

When accessing field data, the script used a single string to specify the name of the field. Recent versions of `yt` now require that we fully specify the two part names of fields. For more context, I provide a concrete example below:
-  we could previously access density data by specifying  `data['density']`. 
- In reality, `yt` didn't directly track a field called `'density'`. Instead it tracked a field called `'enzoe', 'density'` (this field is specific to enzo-e, corresponds to the data stored on disk, and is in code units) and a field called `'gas', 'density'` (this is a field derived from `'enzoe', 'density'` and has cgs units). Previously, `yt` would translate `density` into one of these two choices under-the-hood
- the latest version of `yt` requires that we directly access density by specifying `data['gas','density']` or `data['enzoe','density']`. This removes some ambiguity.

This PR is meant to provide a short-term fix so that these failing tests don't interfere with the review of other PRs (like PR #199).

With that said, the script is a bit of a mess. A lot of complexity arises from the fact that it supports compute the L1 Error Norm when comparing 2 different snapshots and when comparing a snapshot with a tabulated solution (like for a shock tube). Cleaning this script up is definitely on my ToDo list (I also want to migrate the tests using this script to the pytest framework and add the ability to compute the l1 error norm for simulation that use AMR)

### DIFF
--- a/tools/l1_error_norm.py
+++ b/tools/l1_error_norm.py
@@ -312,7 +312,7 @@ def find_common_fields(ds1, ds2, verbose, fields = None):
     else:
         has_field = lambda field: field[1] in ds2
     for field in coerced_fields:
-        if has_field:
+        if has_field(field):
             shared_fields.append(field)
         else:
             missing_fields.append(field)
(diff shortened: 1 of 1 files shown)


gregbryan (approved):
Looks good to me!

jobordner (approved):
Works for me

PR #270, opened 2022-12-06
Add Ability to Configure Floating Point Optimizations and Add Documentation

mabruzzo:
This PR added the ability to configure floating point optimizations and added documentation about this to the getting started page of the webpage. I also adjusted the presentation of other configuration parameters on the webpage documentation.

As an aside, it might make sense to move the description of all of the configuration options to a separate page on the website in order to avoid cluttering up the new-user tutorial.


jwise77 (approved):
Looks good. Thanks for adding to the "getting started" docs!

cmake/EnableFPOptimizations.cmake
```diff
@@ -0,0 +1,155 @@
+# See LICENSE_ENZO file for license and copyright information
+
+#.rst:
+# EnableFPOptimizations
+# ---------------------
+#
+# Defines a function that enables value-unsafe floating point optimization
+# flags for the C and C++ compilers.
+# 
+# This can also adds the openmp-simd flags to the C and C++ compilers (and any
```
gregbryan, line 10:
```suggestion
# This can also add the openmp-simd flags to the C and C++ compilers (and any
```

gregbryan (approved):

PR #273, opened 2022-12-22
Added developer documentation about standardized Method properties ma…

mabruzzo:
This adds separate documentation to the developer's guide unrelated to the material described in PR #265.  (I didn't update the existing PR since John Wise had already reviewed that one)

Specifically, this describes some operations that Cello performs automatically for the developer. The main point is to document behavior related to the Courant factor. It also starts to document some other stuff.


doc/source/devel/method.rst
```diff
@@ -15,7 +15,57 @@ methods that concrete classes override are shown below:
...
    This page is very incomplete. Among other things, we have not discussed
    ``pup`` routines.
+
+
+Standard properties tracked in base class
+=========================================
+
+All ``Method`` classes provide some standardized properties that are managed through the base class.
+
+In some of the following cases, we will talk about how the parameter get's specified for a hypothetical method called ``"my_method"`` (in this hypothetical scenario, subclass's implementation of :cpp:func:`~Method::name` would return ``"my_method"``).
```
gregbryan, line 29:
```suggestion
In some of the following cases, we will talk about how the parameter gets specified for a hypothetical method called ``"my_method"`` (in this hypothetical scenario, subclass's implementation of :cpp:func:`~Method::name` would return ``"my_method"``).
```

doc/source/param/method.incl
```diff
@@ -50,8 +50,13 @@ list is applied in the order specified.  Possible methods include:`
         }
      }
 
+:e:`There are a subset of parameters that can be specified for all
+parameters.  For example, a` :ref:`schedule_param` :e:`subgroup can
```
gregbryan, line 54:
```suggestion
methods.  For example, a` :ref:`schedule_param` :e:`subgroup can
```
(not sure this edit is correct)

gregbryan (approved):
Looks good to me.

jobordner (approved):
Looks good, I just have some comments relating to possible future improvements. Having separate entries for `Method:<foo>:courant` documentation repeats text; factoring that out by moving it up under `Method:courant` might be helpful at some point, leaving specific `Method:<foo>:courant` entries for documenting idiosynchrasies. Renaming courants to "`courant_global`" and "`courant_method`" (or `_local`, or `_extra`) might help reduce confusion. Also I'd like to keep 1.0 as a usable default for method-specific courant factors if possible, which may mean reducing `timestep()` accordingly (and perhaps explicity allowing method-specific courant factors > 1.0).

mabruzzo:
@jobordner  I'm just getting to this now. I have a few ideas that I want to discuss and I think it would be better to do this in a separate issue. I'll merge this PR after I open this issue.

PR #275, opened 2022-12-30
Adding missing grackle chemistry pup's

jobordner:
Recent changes to Grackle (changeset 5547acd on Dec. 2) added a couple of attributes to the `chemistry_data` struct that broke Grackle-related circleci tests in Enzo-E. This PR adds the new attributes to the `EnzoConfig::pup()` method (`max_iterations` and `exit_after_iterations_exceeded`), as well as a couple others that I discovered were missing (`dust_recombination_cooling` and `H2_custom_shielding`).

I had previously sorted the attributes in pup(), but have reverted the ordering to that in `grackle_chemistry_data.h` in Grackle to facilitate code reviews.

### DIFF
--- a/src/Enzo/enzo_EnzoConfig.hpp
+++ b/src/Enzo/enzo_EnzoConfig.hpp
@@ -38,54 +38,62 @@ inline void operator|(PUP::er &p, chemistry_data &c){
     c.grackle_data_file = NULL;
   }
 
-  // [sorted]
-  p | c.bremsstrahlung_cooling_rates;
-  p | c.CaseBRecombination;
-  p | c.cie_cooling;
-  p | c.cloudy_electron_fraction_factor;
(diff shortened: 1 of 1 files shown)


gregbryan (approved):

mabruzzo:
Looks good to me! (It's not obvious to me how this caused things to fail - but I'm all for merging this in).

As an aside, I have a pending PR for the grackle repository, grackle-project/grackle#130, that will allow us to avoid this type of issue in the future (that PR  provides dynamic dictionary-like access to the fields in the `chemistry_data` struct - this will make serialization a lot simpler)

mabruzzo (approved):

jobordner:
@mabruzzo The pup() method is used to broadcast the EnzoConfig attributes to all processes at startup, so if any attributes are missing they will not be initialized on pe's > 0.

Issue #260, opened 2022-07-18
Large (~512^3) cosmology simulations with Grackle failing at cycle 0


WillHicks96:
I ran into this when trying to start a 512^3 simulation on 256 nodes on Frontera with nine-species chemistry out of the current version of `main`. I've recently run smaller simulations (e.g. 128^3) with the same setup without any issues, so it's only happening on larger sims.

The error output suggests that some of the rates being passed into Grackle in `initialize_grackle_chemistry_data` are NaNs. This is somehow related to PR #173. To confirm this, I undid the changes to the relevant files that PR #173 made, and the simulation started without any issues. It did eventually fail with a negative energy error in PPM after ~500 cycles, although I'm not sure if this is a separate issue or if I just screwed something up when undoing the changes in PR #173 (I saved this branch here: https://github.com/WillHicks96/enzo-e/tree/old_grackle).

Full output from the sim: [N512_stdout.txt](https://github.com/enzo-project/enzo-e/files/9135591/N512_stdout.txt)
`parameters.out` file (renamed parameters.txt so I can attach it): [parameters.txt](https://github.com/enzo-project/enzo-e/files/9135588/parameters.txt)
MUSIC parameter file for generating ICs: [N512_CSA.txt](https://github.com/enzo-project/enzo-e/files/9135606/N512_CSA.txt)

mabruzzo:
Sorry it took me so long to get to this, I've started working on this, but If you still have them, would you mind providing me with the versions of the parameter file for the 128^3 run (or at least edit provide a version of the MUSIC parameter file to generate appropriate ICs...)?

mabruzzo:
Sorry for taking so long to address this. This was a rather tricky issue to debug

I _think_ that the issue was that we weren't PUPing the `chemistry_data` struct's `dust_recombination_cooling` and `H2_custom_shielding` fields. Presumably some other differences between the branches caused those parameters to get initialized to different junk values when the parameters got communicated between nodes.

I'm currently running some tests to verify my conclusion, but initial results are promising! (Things no longer crashing)! I'll have an update for you about this later today.

As an aside, I have plans to introduce a PR to completely avoid this class of bugs (in the PUP routine) in the future. It will make use of new functions that I wrote for Grackle (they actually just got merged into Grackle this morning)

mabruzzo:
Following up, I confirmed that after adding the missing pup commands:
- Will's branch where he reverted grackle
[...]

PR #278, opened 2023-01-21
Introduce some debugging functions to dump arrays to disk

mabruzzo:
This is a small PR with some changes to introduce some basic functionality to help me finish resolving PR #260 (I hope to finish resolving that Issue this coming week).

Specifically, I added the functions: `disk_utils::dump_arrays_to_hdf5` and `disk_utils::dump_array_to_hdf5` to let us write arbitrary`CelloArray`s to disk at any time (i.e. in the middle of a method's execution). This is something I've we wished that we had multiple times in the past when I debugged other issues.


src/Cello/disk_utils.hpp
```diff
@@ -0,0 +1,117 @@
...
+
+namespace disk_utils{
+
+
+namespace detail{ // helper functions not meant to be used elsewhere
+
+/// helper function that actually writes ``arr`` to an open HDF5 file
+///
+/// This assumes that ``file`` has been opened and that the appropriate group
+/// has been created (and is openned)
```
gregbryan, line 21:
```suggestion
/// has been created (and has already been opened)
```

gregbryan (approved):
Looks good to me!

mabruzzo:
At some point in the last few months, I became aware that the `typeid` operation employs RTTI (runtime type inference) - which can apparently introduce runtime overhead. Thus, I decided to refactor ``cello::get_type_enum``. The function now operates at compile-time (by default it will provide a useful compile-time error if you try to give a type without a type_enum counterpart).

It was definitely possible to implement this in C++11. But, I realized that the implementation got a lot more concise (and easier to understand) in C++17. So I decided to do that instead. Consequently, this shouldn't get merged until after #325

bwoshea (approved):
Looks good to me!

PR #280, opened 2023-01-29
Reorganize `src/Enzo`

mabruzzo:
### Overview and Motivation

This PR reorganizes the source code files in `src/Enzo`. I think it's really important that we establish the precedent the files in this subdirectory are logically organized into subdirectories.

I think that organizing code into logically related subdirectory is an important change for the following (related) reasons:
1. it will make searching through the repository for code relevant to a some functionality (e.g. for a hydro solver or for the gravity solver) much easier for new developers/users (and for older users/developers). The flat-file organization makes it really hard to know where to start looking for things that you aren't familiar with the code.[^1]
2. it should reduce some hesitance that developers may have about introducing classes or functions that don't directly inherit from the Cello class-hierarchy.[^2] 
3. at least in some cases, it provides boundaries that can be used in some cases to dramatically reduce compile times (e.g. by making subcomponents).[^3] I've left this to be considered in the future on a case-by-case basis

I also want to mention that there is some urgency to getting this done soon - I think the window for reorganizing the codebase before we pick up more users/developers is closing (and as I understand it, that's ultimately why the reorganization of the original enzo codebase failed).

I realize that in this PR I have probably not come up with the optimal organizational structure for the codebase, but I do think that it's a reasonable first approximation. While I'm obviously happy to make changes based on feedback, I would argue that we don't need to stress about getting this **exactly right** on the first pass. We can always move things around later (moving around a few files from time-to-time in the future should not be too painful)

### New Organization

Most files from ``src/Enzo`` have been moved into one of the following new subdirectories:
- `amr`: contains code for refinement criteria, prolongation, and restriction
- `assorted`: mostly uncategorized code that might eventually warrant it's own top level
- `cosmology`: code related to cosmology
- `enzo-core`: code related to core Enzo-E functionality. A lot of the files in this directory are subclasses of Cello classes
- `fluid-props`: code for classes encapsulate fluid properties that code inside and outside of the hydro-solver need to be aware of (e.g. equation of state, whether the dual energy formalism is in use, floors on different fields)
- `gravity`: code related to the gravity solver
- `hydro-mhd`: code related to implementing hydro & mhd integrators
- `initial`: all of the initializers
- `particle`: code related to particles
- `utils`: miscellaneous cross-cutting code that is used in different contexts

The main directory just holds ``enzo-e.cpp`` (which contains the equivalent of the ``main`` function) and a bunch of general headers. We may be able to reduce the number of headers in the main directory, but I'm deferring that for a future PR -- I'm trying to mitigate merge conflicts by avoiding edits to header/source files other than ``_enzo.hpp``. 

<details> 
  <summary> This spoiler tag describes some ideas for reducing headers in the source directory in the future</summary>

  - move some headers into existing directories. For example, it probably makes sense to split the Fortran function declarations in ``enzo_fortran.hpp`` into separate headers and put those headers next to the corresponding fortran code. It might also make sense to put ``enzo_reductions.hpp`` and ``enzo_finalize.hpp`` into ``enzo-core``.
  - consolidate some headers (like ``fortran.h`` with ``fortran_types.h`` or ``charm_enzo.hpp`` with something else)
  - create a directory for some common headers (e.g. ``enzo_constants.hpp``, ``enzo_defines.hpp``, ``enzo_typedefs.hpp``, the fortran headers)
 </details>

Overall, I feel fairly strongly that there should indeed by a ``hydro-mhd`` directory. Currently, my thinking is that should just include stuff in that directly related to the hydro/mhd integrators. At the moment, there is also some functionality related to source terms in that directory, but it may make sense to eventually move those elsewhere.

I think most of the other directories are fairly sensible since fairly clear lines can be drawn between the contents of each of them. I also tend to think that having an ``enzo-core`` subdirectory makes sense (if we don't have a catch-all directory, then ``src/Enzo`` will double in size). But maybe we should move the stuff in ``assorted`` into ``enzo-core``?

I'm definitely open to feedback.

### Alternative Organization strategy:

I discuss an alternative organization strategy in the following spoiler-tag:

<details> 
  <summary><b>Organize-by-related-functionality vs. Organize-by-name</b> </summary>

   A while back somebody had mentioned that they thought it would be better to organize the codebase based on keeping things with similar names together. For example all ``Method`` subclasses would be placed in an ``EnzoMethod`` directory and all ``Compute`` objects would go into a ``EnzoCompute`` subdirectory.

  The name-based approach does indeed clean up the ``src/Enzo`` directory compared to what it looks like now. However, I would argue that this approach doesn't do as good of a job and it won't scale as well in the future since there are plenty of files containing code that aren't currently part of a class hierarchy and never will (examples include ``EnzoBlock``, ``EnzoBoundary``, ``EnzoCenteredFieldRegistry``, ``EnzoDualEnergyConfig``, ``EnzoFactory``, ``EnzoIntegrationQuanUpdate``). Plus, more of these files will appear over time.

  More importantly, the name-based approach:

  - isn't as clear about helping new-users/developers identify files that contain related code 
  - wouldn't help encourage new developers to write classes/functions outside of the standard Cello class hierarchy that make their ``Method`` classes simpler (see point # 2 at the top of this PR). If anything, the name-based approach may discourage new developers from doing that in order to avoid adding more files to ``src/Enzo``
  - provides less-natural boundaries for creating sub-components that don't link against everything. This probably isn't important (but it could become more important as the codebase grows)

  Finally, we could always shift to organization strategies in the future if we really don't like the related-functionality approach

</details>

##### 
 
[^1]: From my own experience, I find that the flat file-structure makes enzo-dev **really hard** to navigate. While `src/Enzo` in Enzo-E _only_ contains ~280 files, it is large enough to cause difficulties. In my own experience, I have found it a little difficult to keep track of which classes are used by the gravity solver or the purpose of certain Fortran files (e.g. there are multiple files that do some kind of interpolation, but they are used in different contexts).

[^2]: When implementing a new ``Method`` class, there have been a number of occasions, where it has been apparent that I could abstract out some data/functionality into it's own helper-class (which would simplify the logic in the ``Method`` class), but I've been hesitant to do that because the helper class doesn't inherit from the ``Cello`` class hierarchy (& it might not be obvious to other developers where that helper class is used). Often times, I only make a helper class/struct to facilitate specialization via subclassing OR if I think that the class could have utility in other contexts.

[^3]: For example, there is no reason why a change to a header for a reconstruction algorithm used in a hydro solver should trigger a rebuild of the entire gravity solver or any of the particle-related functionality. As a solution we could make stuff in the ``gravity`` and ``particle`` subdirectories not depend on the headers in the ``hydro-mhd`` subdirectory.


jwise77:
I'm attaching a list of directories in the failed attempt in creating source directories in classic Enzo. This isn't a suggestion, but I'm posting it to trigger ideas.
image

mabruzzo:
So, I think this is ready for final review.

For future reference, yesterday, May 11, 2023, we had a breakout session about this PR (at the Enzo-classic/Enzo-E Workshop) and there were a few takeaways:

  * @jwise77 suggested that we subdivide some of the source files from the ``particle`` subdirectory
  
  * @jobordner some suggestions about he would organize the directory. Some of the ideas I made use of include:

     * renaming ``Enzo/amr`` to ``Enzo/mesh``

     * placing ``EnzoSolver*`` files and ``EnzoMatrix*`` files into their own nested subdirectories called ``solvers`` and ``matrix``

     * creating an ``io`` subdirectory

     * creating an ``obsolete`` subdirectory (we should talk what about what it would take to drop these files completely, but that should be addressed in a future PR)

     * creating a ``tests`` subdirectory for initializers that are soley used for testing purposes. (I'm honestly a little torn about this... Since the line between an initializer used for testing and other initializers could be blurry). But, I don't feel very strongly about this


A lot of @jobordner's suggestions were fairly similar to mine. I just wanted to highlight some differences and the my rationale for some of my choices:

  * I have left a bunch of header files in the main src directory for this now (moving these around might require some header modifications, and that's probably easier to handle separately)

  * I tried to avoid "over-decomposing" files into too many directories. I think my basic rule-of-thumb was I needed at least 6 source/header files to warrant a subdirectory. (This isn't intended as a hard-and-fast rule, but it was just a place to start). I wanted to highlight a subset of @jobordner's suggestions that I didn't follow for this reason, but I think may be excellent ideas for the future (as we add more files):

     * ``Enzo/control`` which could contain ``EnzoStopping.[ch]pp``

     * ``Enzo/parallel`` which could contain ``EnzoMethodBalance.[ch]pp`` and ``EnzoMsgCheck.[ch]pp``

     * ``Enzo/parameters`` which could contain ``EnzoConfig.[ch]pp``

  * I opted to create ``gravity/solvers`` and ``gravity/matrix`` rather than ``numerical/solvers`` and ``numerical/matrix`` since the gravity solver is currently the only thing that makes use of this multigrid functionality (but if we ever add other physics-solvers that use multigrid functionality, I think this is an excellent idea)

  * the organization of the particle-stuff is different. I thought the it might make more sense to put helper routines for accessing particle data next to the physics methods that actually use them. But, we can definitely revisit this in the future (this choice is mostly from inertia)

  * I choose not to put the various physics-methods into a physics subdirectory (e.g. we could make ``physics/cosmology``, ``physics/hydro-mhd``, ``physics/gravity``). Since these include such a large fraction of the Enzo-E files, I think it makes a similar level of sense to leave these at the root-level of the ``Enzo`` directory (But, it could help make more sense for organizing the files ins ``assorted``). The main reason for my hesitation is coming up with a good name.

     * I'm also a little worried about confusion with the ``Physics`` class hierarchy.

     * @jwise77's screenshot shows that during Enzo-classic's reorganization they made a similar directory called solvers (But that could be confused with the unrelated ``Solver`` hierarcy).

     * I think this is something we can easily revisit in the future.

mabruzzo:
Let me know if there's any other changes that you want to see. Otherwise I'd suggest that we pull the trigger (we can always fix things later)

src/Enzo/obsolete/CMakeLists.txt
gregbryan, line 1: Do we really want to compile code in the absolute directory?

gregbryan (approved):
Looks good -- just one question.   Thanks!

src/Enzo/obsolete/CMakeLists.txt
mabruzzo: I think we will plow ahead and I will open another PR next week to remove the obsolete directory

Issue #211, opened 2022-06-13
Setting zlib path


setsuna0402:
I met an error about that cmake confuses which zlib should be used, if there are multiple zlibs in the system. (hdf5 library depends on zlib.)   
One solution is to allow users clarify the path of zlib in config/*.cmake files.

mabruzzo:
As we discussed offline, the problems arise when you try to use a custom HDF5 installation that depends on a custom zlib installation.

I dug into this last week. It sounds like your `HDF5_C_LIRARIES` and/or `HDF5_C_INCLUDE_DIRS` variables are broken in your build. These variables are determined based on the output of `/path/to/h5cc -show -shlib`. This suggests that either:
- your custom HDF5 library's `h5cc` compiler wasn't exactly correct. (Would you mind sharing the output of the above command?)
- there is a bug in the FindHDF5 module that ships with CMake

In any case, it would probably be nice to provide an escape hatch for people to specify extra linking directories or extra libraries (a similar solution seems to be necessary to address an issue @wolfram-schmidt encountered with missing libraries while trying to use the intel compilers)

The HDF5  and the zlib dependency is picked when `

setsuna0402:
Hi,  I just checked the path of zlib used by my hdf5 library. It is correct. So the reason may be the later one.

mabruzzo:
I saw something earlier today that made me realize that the pngwriter library may also link to zlib (with may have some relationship with libpng), which further complicates the issue

PR #281, opened 2023-02-03
Isolate hdf5 as a dependency of Cello's disk component 

mabruzzo:
### Overview

Slightly tweaked Enzo-E so that `libhdf5` is now a private dependency of Cello's disk component. Consequently, libhdf5 (and it's own transitive dependencies) are no longer a transitive dependency of the rest of the codebase (that depend on the disk component).

### Motivation

This was done as a step towards making problems like issue #211 easier to resolve 
- they were having trouble linking against the correct version of `zlib`. For context, `zlib` is a transitive dependency of both `libhdf5` and `libpng`. 
- From in-person conversations with that user, I think that the they were using a manual installation of hdf5 that depended upon a manual installation of `zlib`, while I think that their `libpng` may have relied on the system level installation of `zlib`.
- in the future, I plan to make `libpng` a private dependency of another Cello subtarget. Even if this doesn't completely solve the problem, it should make this sort of problem a lot easier to diagnose in the future...

### Details

I accomplished this by hiding the `FileHdf5` class from the rest of the codebase (`disk_FileHdf5.hpp` is no longer included in the global headers). `FileHdf5` is now constructed by the new static method of the `File` class called `File::construct_FileHdf5`, which returns `File*`.

I did want to highlight another approach for making `libhdf5` a private dependency: the pimpl idiom (described [here](https://en.cppreference.com/w/cpp/language/pimpl) and [here](https://en.wikipedia.org/wiki/Opaque_pointer)). I've also heard that approach described as defining an opaque pointer or using the bridge design pattern. It may make more sense to refactor `FileHdf5` to employ that idiom and delete the `File` base class altogether (it seems unlikely that we would define another subclass for `File`). @jobordner please let me know if you think that would be better.


src/Cello/disk_File.hpp
```diff
@@ -18,6 +18,9 @@ class File {
 
 public: // interface
 
+  /// factory method for constructing a FileHdf5 instance
```
gregbryan, line 21: Maybe useful to explain briefly here why this is implemented as a factory method.

gregbryan (approved):
Seems sensible to me -- an additional advantage of this approach is that compilation will be marginally faster with less baggage that needs to be included into every source file compile.

jobordner (commented):
The same effect can be obtained much more simply by 1. relocating `#include "hdf5.h"` from `_disk.hpp` to `disk_FileHdf5.cpp`, 2. replacing `hid_t` in `FileHdf5` method parameters and return types with `int64_t`  or equivalent (and 3. removing the unused `FileHdf5::data_dims_` attribute.) No need for structural changes.

mabruzzo:
Good point. I'll take a stab at that

PR #283, opened 2023-02-05
``CelloArray`` to ``CelloView`` rename

mabruzzo:
This renames ``CelloArray`` to ``CelloView``. It also changes the name of Cello's array component so that it is now called the view component.

This change reflects the fact that ``CelloArray`` doesn't have semantics like C++ arrays. It really has semantics that more closely mirror a smart pointer or a "view" (views are new in C++ 20). The documentation provides a brief discussion about this. Moreover, its semantics that are similar to a ``Kokkos::View`` or a ``RAJA::View``


jobordner (approved):
I'd consider removing the `using` "backward-compatibility" line in `_view.hpp`, otherwise `CelloArray` could creep back into the code. Otherwise looks good.

gregbryan (approved):
Looks good to me! A few very minor typo suggestions...

mabruzzo:
> I'd consider removing the `using` "backward-compatibility" line in `_view.hpp`, otherwise `CelloArray` could creep back into the code. Otherwise looks good.

Good point. I just made that change

Issue #247, opened 2022-07-01
License Incompatibility: PNGwriter


mabruzzo:
Cello/Enzo-E currently packages (and links against) the PNGwriter library (see the contents of `src/External`. That library is released version 2 of the GPL license.

This is incompatible with Cello/Enzo-E's license. We really need to deal with this now that we approaching maturity.

stefan-hdt:
Hi there, 

the GPL-2.0 does not seem (completly) incompatible with the Cello/Enzgo-E's license. At least you could publish your source code in combination with PNGwriter unter the GPL-2.0.

In case you do not want to use a OSS license with a strong copyleft effect, there is only two ways for you to deal with the situation: (1) Use a different or alternative library than PNGwriter or (2) feel happy with the fact that your work will be available for others under Cello/Enzo-E's license and only the combination with the dependency PNGwriter will trigger the copyleft effect.

Hope I could help a little bit.

Kind regards,

Stefan

PR #284, opened 2023-02-07
PNG-Routine Consolidation

mabruzzo:
Previously, PNG routines were used in 4 different locations throughout the codebase. These routines were used to read a png file as a mask OR to write data to a png file.

This PR defines 2 functions in ``pngio::write`` and ``pngio::read_as_mask`` in the disk component, and refactors those other areas of the codebase to directly call these functions. 

This involved simplifying code in the ``OutputImage`` class. In practice, the actual operations completed in the revised version of ``OutputImage`` are essentially the same (the only exception is that we perform an extra heap allocation and an extra copy in 1 special circumstance).

The benefits of these changes are two-fold:
1. It will make excising of ``pngwriter`` from the codebase a lot easier (see Issue #247) - since all code involving that library is located in 1 place
2. It allows us to isolate the PNG-related dependencies to a single source file. They are no longer a transitive dependency of other components (this will help resolve issues like #211)

My naming of things is pretty arbitrary, so definitely let me know if you want me to change anything


src/Cello/disk_pngio.hpp
```diff
@@ -0,0 +1,45 @@
...
+  ///    the range of values that the colormap covers (after the transform has
+  ///    been applied to entries of data). Values outside of this range are
+  ///    clamped to the nearest value. Pass ``nullptr`` to indicate that the
+  ///    full range of elements in data should be mapped to colors.
+  void write(const std::string& fname, double* values, int width, int height,
+             const std::vector<float> (&colormap)[3],
+             ImgTransform transform,
+             const std::array<double,2>* min_max) noexcept;
+
+  /// reads data from a png file to create a mask
```
gregbryan, line 35: Maybe add comment to indicate that mask is true if any RGB channel in png image is non-zero?

src/Cello/io_OutputImage.hpp
```diff
@@ -103,9 +102,28 @@ class OutputImage : public Output {
   virtual void init () throw();
 
   /// Open (or create) a file for IO
-  virtual void open () throw();
+  ///
+  /// @note
+  /// In practice, this method performs a no-op. It's mostly useful in
+  /// situations where we can write to a file in chunks. However, the current
+  /// implementation constructs the image in memory and the writes the image
```
gregbryan, line 109:
```suggestion
  /// implementation constructs the image in memory and then writes the image
```

gregbryan (approved):
Looks good to me.

PR #285, opened 2023-02-26
Nested initial grids

brenjohn:
### Overview
Added some infrastructure to facilitate nested initial grids: 
 - Member functions were added to Cello and EnzoE Block classes. These can be used to check if a block should refine during the initialization phase and to create the child blocks if it should.
 - Helper classes were introduced to the EnzoInitialHdf5 files called DataLoaders. These were introduced to facilitate reading HDF5 data for blocks on different levels and to avoid introducing further code duplication.
 - The Config and Hierarchy classes were updated to read and store parameters from the parameter file regarding nested initial grids. 

### Details
The following changes were made to the input parameter file to read in nested initial grids:
 - A level parameter was introduced to the file groups within the HDF5 group to indicate which level the HDF5 file contains data for.
 - Lower and Upper parameters where added to the Mesh group for each level of the initial grid. These define a region of parent blocks on level `L-1` which need to refine to create the nested grid at level `L`.
 - The `max_level` parameter in the HDF5 group is used to only initialize levels up to level `max_level`.
 - The member function `index_from_global` was added to the Block class which returns the index of a block corresponding to given global coordinates of the block.


fearmayo (approved):

mabruzzo:
If it's not too much trouble, it probably makes sense to do a rebase so that the ics aren't in the commit history (let me know if you need some help with how to do that).

If we choose not to do that, then the ics will inflate the size of the repository as we move forward.

brenjohn:
You read my mind! :P 

I was just looking into ways to remove the ics from the commit history. I also planned on doing a rebase anyway to take in new changes from main. I'll be in touch if I get stuck.

I was thinking the same about an enzo-project repository for files like these and was going to bring it up in the next meeting. I don't think what I've done here is ideal long term. I was thinking maybe a data directory in the top level of enzo-e which we could turn into a git submodule might make sense but we should discuss this first and would go in its own PR.

.circleci/config.yml
jobordner, line 1: The IC's for the test problem should probably be brought into the input directory to avoid referencing external projects.

doc/source/param/mesh.incl
jobordner, line 1: To help avoid confusion about "level_n_lower" and "level_n_upper" being the literal parameter names, I'd update the markup to something like
```
:Parameter:  :p:`Mesh` : :p:`level_`:g:`<n>`:p:`_lower`
```
A short example could also be helpful.

src/Cello/mesh_Block.cpp
jobordner, line 1: `index_from_global(ix,iy,iz, nx,ny,nz, ax,ay,az)` can be removed since it's never called

jobordner (changes requested):
Overall this looks great! 

While I don't expect it to be used for 2D problems, Enzo-E does explicitly support rank < 3, so I would add something to the documentation stating this is implemented for 3D only, and maybe add an ASSERT to check that.

 I also didn't see a reference to cycle in `refine_during_initialization()`, which I'm guessing could be an issue if for example the initialially refined blocks are coarsened in some later cycle(?)

Nice work!

brenjohn:
Thanks @jobordner

I think I've addressed everything now. Let me know if there's anything else.

Just regarding the comment about initial blocks possibly coarsening at a later cycle. From what I understand this would be intended behavior and we want blocks to coarsen if they can, even if they were part of an initially refined region. The increased particle resolution should still be there, possibly in some other block.

jwise77:
@jobordner Can you review John B's recent changes that addressed your comments? If all looks good, please go ahead and approve. Thanks!

jobordner (approved):

PR #288, opened 2023-03-06
More tweaks to the parameter component

mabruzzo:
This PR builds on #287 and introduces some additional tweaks to the parameter component. Among other things, it addresses a bug in `ParamNode::pup`.

This has been separated from #287 to make #287 as small as possible (so that it can be reviewed quickly). The other changes in this PR aren't as urgent


gregbryan (approved):
Looks good to me but @jobordner should probably review.

PR #293, opened 2023-03-11
Removing static variables from EnzoBlock

mabruzzo:
Removed remaining static variables from ``EnzoBlock``. These variables all either:
- held redundant information that was already stored elsewhere
- hold configuration parameters for the ppm solver.

I simply removed attributes in the former case and changed attributes in the latter case so that they are now stored as attributes of `EnzoMethodPpm`


src/Enzo/hydro-mhd/SetMinimumSupport.cpp
```diff
@@ -12,10 +12,15 @@
 //----------------------------------------------------------------------
  
 int EnzoBlock::SetMinimumSupport(enzo_float &MinimumSupportEnergyCoefficient,
+                                 enzo_float minimum_pressure_support_parameter,
 				 bool comoving_coordinates)
 {
   const int in = cello::index_static();
```
gregbryan, line 18: Is this line required at all anymore?

src/Enzo/hydro-mhd/EnzoMethodPpm.cpp
```diff
@@ -265,7 +288,7 @@ double EnzoMethodPpm::timestep ( Block * block ) throw()
 			&enzo_block->CellWidth[0],
 			&enzo_block->CellWidth[1],
 			&enzo_block->CellWidth[2],
-			&gamma, &EnzoBlock::PressureFree[in], &cosmo_a,
+			&gamma, &pressure_free_int, &cosmo_a,
```
gregbryan, line 291: I think this change means you can remove the line setting `in`

gregbryan (approved):
This PR looks good to me.

src/Enzo/hydro-mhd/EnzoMethodPpm.cpp
mabruzzo: good catch! I removed it!

src/Enzo/hydro-mhd/SetMinimumSupport.cpp
mabruzzo: This argument is used  is used inside of `SetMinimumSupport`. Specifically, the argument is multiplied all of the other variables that are used to compute `MinimumSupportEnergyCoefficient`

gregbryan (approved):
Approved!

PR #272, opened 2022-12-19
Added a primer on CMake to docs and tweak source file search

mabruzzo:
I've been thinking about writing the "CMake Primer" ever since the last developer's workshop. Please let me know whether you think this is useful and if there is anything that doesn't make sense. Also let me know if you think I should add to anything else

The tweak to the source file search should makes the build system re-execute the globbing expressions whenever you execute the build-system. If the detected files change, the build system gets updated. This should reduce the frequency with which you need to start builds from scratch.

PR #298, opened 2023-03-23
Suppress warnings charm++ module headers

mabruzzo:
This PR builds directly on PR #272 (it shouldn't be merged until afterwards).

This PR takes steps to suppress compiler warnings in the header files generated by ``charmc`` from ``.ci`` (these warnings are really just noise -- we can't do much about them without altering charm++). 

This is essentially done by telling the compiler to treat them like system headers with the (i.e. passing the `-isystem` flag to `gcc` or `clang` -- this may not suppress warnings on all compilers, but it's a good start)

In more detail, we now:
1. place the headers generated by ``charmc`` from ``.ci`` files into their own sub-directories -- a subdirectory is automatically made for headers generated from the ``ci`` files from the ``Cello``-layer and another is created for the headers generated from the ``ci`` file from the ``Enzo``-layer.
2. label all headers in these subdirectories as system-headers.


cmake/charm.cmake
```diff
@@ -84,26 +85,70 @@ set(__addCharmModule YES)
...
+# Suppressing Compiler Warnings
+# =============================
+# Headers generated by certain versions of Charm++ have been found to trigger
+# compiler warnings (e.g. gcc's -Wsign-compare and -Wunused-variable). Given
+# that applications have no control over these warnings, it's desirable for
+# these warnings to be suppressed. This is done by declaring the header files
+# generated by Charm++ to be system header files.
+#
+# If you would prefer to not declare them as system headers, define the
+# ``SUPRESS_CHARM_HEADER_WARNING`` global variable with a value of ``0``.
```
gregbryan, line 110:
```suggestion
# ``SUPPRESS_CHARM_HEADER_WARNING`` global variable with a value of ``0``.
```

cmake/charm.cmake
```diff
@@ -84,26 +85,70 @@ set(__addCharmModule YES)
...
 
   add_library(${MODULE}CharmModule INTERFACE)
   add_dependencies(${MODULE}CharmModule ${MODULE}CharmModule_custom_target_)
 
-  # all generated headers are placed into ${CMAKE_CURRENT_BINARY_DIR}
-  target_include_directories(${MODULE}CharmModule
-    INTERFACE ${CMAKE_CURRENT_BINARY_DIR})
+  # The following command informs all targets that will be linked against
+  # ${MODULE}CharmModule where to find the generated headers
+  if(SUPRESS_CHARM_HEADER_WARNING OR NOT DEFINED SUPRESS_CHARM_HEADER_WARNING)
```
gregbryan, line 148:
```suggestion
  if(SUPPRESS_CHARM_HEADER_WARNING OR NOT DEFINED SUPPRESS_CHARM_HEADER_WARNING)
```

gregbryan (changes requested):
A couple of typos that need to be fixed, otherwise good.

cmake/charm.cmake
mabruzzo: Fixed

cmake/charm.cmake
mabruzzo: Fixed

mabruzzo:
This is one of the PRs where we decided that we only needed 1 approval and Greg signed off on it (even though he never explicitly hit approve), so I'm merging.

PR #300, opened 2023-03-26
``Value`` class tweaks

mabruzzo:
This PR tweaks the ``Value`` class (and consequently the ``ScalarExpr`` class), so that it's easier to work with instances of the class more easily (rather than always working with pointers to instances). These changes mainly involve tweaks to the constructors/destructors.

It also adds a minor optimization to ``ScalarExpr`` (relevant for inflow boundaries).

Finally, the ``test_Value`` unit tests were refactored and extended to explicitly test this new functionality.


gregbryan (approved):

src/Cello/problem_ScalarExpr.cpp
```diff
@@ -70,6 +53,18 @@ void ScalarExpr::evaluate (T * value, double t,
 	  ndx,ndy,ndz,nx,ny,nz,
 	  (ndx >= nx) && (ndy >= ny) && (ndz >= nz));
 
+  if ((!param_) && (!mask)) {
+    // fast-path that avoid heap allocations - relevant for inflow conditions
+    for (int ix=0; ix<nx; ix++) {
+      for (int iy=0; iy<ny; iy++) {
+	for (int iz=0; iz<nz; iz++) {
+          value[ix + ndx*(iy + ndy*iz)] = (T)value_;
```
jobordner, line 61: Since arrays are stored using column-major ordering I try to order loops with iz on the outside and ix on the inside. This leads to more sequential memory accesses, which can improve cache performance.

src/Cello/problem_ScalarExpr.cpp
```diff
@@ -155,3 +150,15 @@ template void ScalarExpr::evaluate
...
 
+//----------------------------------------------------------------------
+
+std::string ScalarExpr::expr_to_string() const throw()
+{
+  if (param_){
+    return param_->value_to_string(param_write_monitor);
+  } else {
+    char buffer[25] = "";
+    sprintf(buffer, "%#.16g", value_);
```
jobordner, line 161: Shouldn't be a problem, but snprintf is safer.

jobordner (approved):
Looks good, just a couple suggestions.

src/Cello/problem_ScalarExpr.cpp
mabruzzo: Fixed

src/Cello/problem_ScalarExpr.cpp
mabruzzo: Fixed - I though it was weird that I ordered the loop in this way, so I looked around.

It turns out that there were some other pre-existing code in that file that ordered loops in this way. I fixed those loops too.

PR #302, opened 2023-04-10
Unifying and elevating EOS functionality in Enzo-E

mabruzzo:
This PR finally introduces some changes that I've been working towards for about 2 years now.

This PR does 2 things:
1. It makes the EOS a global property of the simulation that is configured through the ``"fluid_props"`` physics object (previously we implicitly assumed an ideal-gas eos everywhere other than the ppml solver). 
    - The main point is to try to have a single source of truth where various EOS-related functionality is implemented in the C++ codebase.
    - The secondary goal is to make it easy for various Methods to either 
        - easily state their assumptions what kind of EOS is used (and cause the code to exit with an error when that is violated)
        - customize their behavior based on the specified type of EOS

2. Previously the VL+CT solver had implemented 2 disconnected versions of the EOS. This PR essentially removed the version with the class-based inheritance. Now we use the version that was employed within the Riemann Solvers everywhere.

~~This PR is 95% done. I really need just need to finish writing up the documentation for this. In particular, I think the way that I've opted to implement polymorphism is probably un-intuitive for most people, and the documentation will justify why we adopt this strategy (and explain why it is the superior choice to class-based inheritance).~~

~~Once I've finished with that, I update this PR to indicate it's no longer a WIP~~

EDIT: This is now ready for review. The website has been updated with a bunch of new documentation. We need to update the gold standard to reflect the recent changes (but that needs to be handled after review)


src/Enzo/enzo_EnzoMethodPpml.cpp
```diff
@@ -28,6 +28,14 @@ EnzoMethodPpml::EnzoMethodPpml()
   const EnzoDualEnergyConfig& de_config = fluid_props->dual_energy_config();
   ASSERT("EnzoMethodPpml::EnzoMethodPpml",
          "incompatible with dual energy formalism", de_config.is_disabled());
+
+  // technically, the PPM solver doesn't directly use any of the functionality
```
bwoshea, line 32: It should probably say PPML instead of PPM in this entire code block, right?

src/Enzo/enzo_EnzoPhysicsFluidProps.cpp
```diff
@@ -1,44 +1,88 @@
...
+
+  // in c++17 this whole function simplifies down to:
+  //
+  // return eos_variant_.visit([](auto eos) -> enzo_float {
+  //   using T = std::decay_t<decltype(eos)>;
+  //   if constexpr (std::is_same<T,EnzoEOSIdeal>::value) { return eos.gamma; }
+  //
+  //   std::string name = T::name();
+  //   ERROR1("EnzoPhysicsFluidProps::gamma()",
+  //          "can't return gamma for eos of type \"%s\". Either that type "
```
bwoshea, line 29:
```suggestion
  //          "can't return gamma for EOS of type \"%s\". Either that type "
```

src/Enzo/enzo_EnzoPhysicsFluidProps.cpp
```diff
@@ -1,44 +1,88 @@
...
+  //   }
+  // });
+
+  if (eos_variant_.holds_alternative<EnzoEOSIdeal>()) {
+    return eos_variant_.get<EnzoEOSIdeal>().gamma;
+  }
+
+  std::string name = eos_variant_.visit(NameVisitor());
+  ERROR1("EnzoPhysicsFluidProps::gamma()",
+         "can't return gamma for eos of type \"%s\". Either that type "
```
bwoshea, line 41:
```suggestion
         "can't return gamma for EOS of type \"%s\". Either that type "
```

src/Enzo/enzo_EnzoPhysicsFluidProps.hpp
```diff
@@ -49,23 +50,114 @@ class EnzoPhysicsFluidProps : public Physics {
...
   { return fluid_floor_config_; }
 
-  enzo_float gamma() const noexcept
-  { return gamma_; }
+  /// Access the EOS
+  const EnzoEOSVariant& eos_variant() const noexcept
+  { return eos_variant_; }
+
+  /// Utility method that tries to retrieve gamma from the stored eos_variant.
+  /// When the stored eos variant doesn't contain a value of gamma, the program
```
bwoshea, line 70:
```suggestion
  /// When the stored EOS variant doesn't contain a value of gamma, the program
```

bwoshea (changes requested):
Overall the code looks great to me and is a good thing to have in Enzo-E.  I have a bunch of very minor suggestions, mostly relating to the documentation and so forth.  I'm happy to approve once you make the requested updates!

dcollins4096 (approved):

src/Enzo/enzo_EnzoMethodPpml.cpp
mabruzzo: You're absolutely right. I'll fix that

doc/source/devel/eos-fluidprops.rst
```diff
@@ -0,0 +1,497 @@
...
+.. math::
+
+  C_0 \times \min\left(\frac{\Delta x}{|c_{s,ijk}+ v_{x, ijk}|},
+                       \frac{\Delta y}{|c_{s,ijk}+ v_{y, ijk}|},
+                       \frac{\Delta z}{|c_{s,ijk}+ v_{z, ijk}|}\right)
+
+where :math:`C_0` is the courant factor (a constant between 0 and 1) and :math:`\Delta x,\, \Delta y,\, \Delta z` specify cell widths.
+
+The following code snippet shows a somewhat simplified example of how you might do this.
+It assumes that the pressure field has been precaluated it will abort if the program does not use an ideal eos:
```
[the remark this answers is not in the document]
mabruzzo: The discussion of the pressure field was very confusing (it was a half-baked thought). I have replaced the second paragraph with the following blurb. Let me know if this addressed your concerns.

```rst
The following code snippet shows a somewhat simplified example of how
you might perform this calculation.  This snippet will abort with an
error if each Processing Element's global
:cpp:class:`!EnzoPhysicsFluidProps` instance was configured to hold
anything other than an ideal EOS.
```

doc/source/devel/eos-fluidprops.rst
```diff
@@ -0,0 +1,497 @@
...
+                   CelloArray<const enzo_float, 3> velocity_y,
+                   CelloArray<const enzo_float, 3> velocity_z,
+                   CelloArray<const enzo_float, 3> pressure,
+                   double dx, double dy, double dz,
+                   double courant_factor)
+   {
+     // the program aborts with an error if Enzo-E was not configured with an
+     // ideal eos (as an aside, we are technically making a copy of the eos
+     // here - that should be totally fine sinces its just a memcpy)
+     const EnzoEOSIdeal eos = enzo::fluid_props()->eos_variant().get<EnzoEOSIdeal>();
```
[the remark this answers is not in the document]
mabruzzo: The short answer is: NOT YET. It's absolutely possible and it was originally a major motivation for introducing the ``visit`` function. However, in practice it's not worth doing right now when we only have two equations of state and while we are using C++11.

Here are 2 examples (while they are possible in C++11, I've opted not to show that right now because that is more cumbersome).

If every EOS class has a method with the signature ``enzo_float EOSClass::sound_speed(enzo_float density, enzo_float pressure)``, then we can write easily write the following snippet in C++14 that will do this:

```c++

   double timestep(CelloArray<const enzo_float, 3> density,
                   CelloArray<const enzo_float, 3> velocity_x,
                   CelloArray<const enzo_float, 3> velocity_y,
                   CelloArray<const enzo_float, 3> velocity_z,
                   CelloArray<const enzo_float, 3> pressure,
                   double dx, double dy, double dz,
                   double courant_factor)
   {
     auto lambda_func = [=](auto eos) -> double
     {
       const int mx = density.shape(2);
       const int my = density.shape(1);
       const int mz = density.shape(0);

       double dt = std::numeric_limits<double>::max();
       for (int iz = 0; iz < mz; iz++) {
         for (int iy = 0; iy < my; iy++) {
           for (int ix = 0; ix < mx; ix++) {

             double cs = (double) eos.sound_speed(density(iz,iy,ix),
                                                  pressure(iz,iy,ix));
             double abs_vx = std::fabs((double) velocity_x(iz,iy,ix));
             double abs_vy = std::fabs((double) velocity_y(iz,iy,ix));
             double abs_vz = std::fabs((double) velocity_z(iz,iy,ix));
             double tmp = std::min(std::min(dx/(abs_vx + cs),
                                            dy/(abs_vy + cs)),
                                   dz/(abs_vz + cs));
             dt = std::min(dt, tmp);
           }
         }
       }
       return dt;
     };
     double dt = enzo::fluid_props()->eos_variant().visit(lambda_func);
     return courant_factor * dt;
    }
```

In practice, it doesn't make sense for every EOS to have a method that computes the sound speed from both density and pressure. In the future, one could imagine modifying ``EnzoEOSIsothermal`` so that it has the method ``enzo_float EnzoEOSIsothermal::sound_speed()``. In C++17 we could write a new function that makes use of that new functionallity:

```c++

   double timestep(CelloArray<const enzo_float, 3> density,
                   CelloArray<const enzo_float, 3> velocity_x,
                   CelloArray<const enzo_float, 3> velocity_y,
                   CelloArray<const enzo_float, 3> velocity_z,
                   CelloArray<const enzo_float, 3> pressure,
                   double dx, double dy, double dz,
                   double courant_factor)
   {
     auto lambda_func = [=](auto eos) -> double
     {
       using T = std::decay_t<decltype(arg)>;
       const int mx = density.shape(2);
       const int my = density.shape(1);
       const int mz = density.shape(0);

       double dt = std::numeric_limits<double>::max();
       for (int iz = 0; iz < mz; iz++) {
         for (int iy = 0; iy < my; iy++) {
           for (int ix = 0; ix < mx; ix++) {

             double cs;

             if constexpr (std::is_same_v<T, EnzoEOSIdeal>) {
               cs = (double) eos.sound_speed(density(iz,iy,ix),
                                             pressure(iz,iy,ix));
             } else if constexpr (std::is_same_v<T, EnzoEOSIsothermal>) {
               cs = (double) eos.sound_speed();
             } else {
               static_assert(false, "can't handle specified EOS");
             }

             double 
             double abs_vx = std::fabs((double) velocity_x(iz,iy,ix));
             double abs_vy = std::fabs((double) velocity_y(iz,iy,ix));
             double abs_vz = std::fabs((double) velocity_z(iz,iy,ix));
             double tmp = std::min(std::min(dx/(abs_vx + cs),
                                            dy/(abs_vy + cs)),
                                   dz/(abs_vz + cs));
             dt = std::min(dt, tmp);
           }
         }
       }
       return dt;
     };
     double dt = enzo::fluid_props()->eos_variant().visit(lambda_func);
     return courant_factor * dt;
    }
```

Note that the if-statements in the for-loop are called "constexpr-if-statements". As I understand it, they will not produce branching within the for-loop; instead separate versions of the lambda expression are automatically generated for each possible EOS type (so effectively there is a branch outside of the if-statement).

doc/source/devel/eos-fluidprops.rst
```diff
@@ -0,0 +1,497 @@
...
+
+    // fetch the value of gamma
+    enzo_float gamma = eos.get_gamma();
+
+    // do work with gamma...
+  }
+
+Now, let's break this down in slightly more detail.
+
+  1. As you're probably aware, :cpp:func:`!enzo::fluid_props()` returns a pointer to the instance of the :cpp:class:`!EnzoPhysicsFluidProps` that is configured for the Processing Element (PE).
```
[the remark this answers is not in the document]
mabruzzo: Gotcha. I've fixed that

mabruzzo:
@bwoshea This is ready for you to take another look at.

The failing single-precision answer test is expected. Once you approve the PR, I will take the necessary steps to update the answer tests reference results

mabruzzo:
Thanks! Merging now!

Issue #138, opened 2022-01-25
Fields are not converted to conservative form when interpolated using EnzoProlong (SecondOrderA)


jobordner:
This is a known issue. Some fields require scaling to convert to conservative form before being interpolated or coarsened, e.g. velocities since it's actually momentum that's conserved. This is done for the simple inter-level transfer operations, but not yet for EnzoProlong. 
The difficulty is that EnzoProlong requires extra communication between neighboring blocks, and the density field is not always available when needed for scaling. (This could be addressed e.g. by always including the density field in the EnzoProlong list of fields to be interpolated.)

PR #303, opened 2023-04-12
More starss fixes

WillHicks96:
This PR adds some minor edits to EnzoMethodFeedbackSTARSS. 

A bug that is addressed here is one in which overlapping SNe bubbles originating from either side of a level boundary would sometimes result in negative values being deposited into the `total_energy` and `internal_energy` fields. The source of this bug is likely related to issue #138. The current behavior for refresh with `accumulate=true` is to revert to linear interpolation when adding ghost zone data to active zones between blocks at different levels. 

Since this is just a rare-ish edge case, I've made the change of forcing a minimum value of zero for values added to the `total_energy` and `internal_energy` fields after the "accumulate" step as a duct-tape solution.

There are also some small modifications to the deposited momentum values corresponding to the different phases of SN remnant evolution -- free expansion, sedov-taylor, and terminal -- to make the method more resolution insensitive.

The attached plots show the results of ideal tests measuring the terminal momentum at various resolutions both before (top) and after (bottom) the edits in this PR. These tests set of a single supernova at the center of the box (on a block corner), and evolve hydrodynamically with six-species gas chemistry. Ideally, this curve would be perfectly straight. Note that before, the terminal momentum in the dx=10-20 pc range were off from the true terminal momentum by a factor of ~3x. Now, the error is smaller at around 1.5x. The terminal momenta at very poor resolutions are correct in both cases. The deviations from the analytical value of $p_\mathrm{term}$ at moderate-high resolutions that persist could be coming from errors in determining where the actual borders for the different supernova phases lie.

pterm_vs_dx_full

pterm_vs_dx_030722

With the choices of density and temperature, the cutoffs between the different phases at deposition are as follows:

- dx < 1 pc -- Free expansion phase
- 1 pc < dx < 6pc -- Sedov-Taylor phase
- 6pc < dx -- Terminal phase
- Analytical $p_\mathrm{term} = 2\times 10^{5} M_\odot km/s$

A parameter file for the test with cell width 5 pc is attached.

[method_feedback_starss_05.txt](https://github.com/enzo-project/enzo-e/files/11215787/method_feedback_starss_05.txt)


src/Enzo/enzo_EnzoMethodFeedbackSTARSS.cpp
```diff
@@ -1053,11 +1025,11 @@ void EnzoMethodFeedbackSTARSS::deposit_feedback (Block * block,
   double pTerminal;
   if (Z_Zsun > 0.01) {
     // Thornton, 1998, M_s * V_s, eqs 22, 23, 33, 34
-    pTerminal = 1.6272e5 * pow(ejectaEnergy/1e51, 13./14.) * pow(n_mean, -0.25) * pow(Z_Zsun, -0.36);
+    pTerminal = 2.7298e5 * pow(ejectaEnergy/1e51, 13./14.) * pow(n_mean, -0.12) * pow(Z_Zsun, -0.14); // Msun km/s
```
jwise77, line 1028: Where did the changes in normalization and slope come from in `p_free` and `pTerminal`?  Are they different cases in Thorton (1998)? Also I couldn't find this paper / book.

jwise77 (commented):
All looks good! I just had one clarifying question about the changes in pressure expressions.

jwise77 (approved):

bwoshea (commented):
@WillHicks96 The PR itself seems fine to me, but (1) the PR needs to be merged with the tip of the main branch, and (2) the tests are failing due to a Grackle issue.  Can you address these?  I'm also going to check on Slack to see if people know how to fix the Grackle problem - I vaguely recall this happening in the recent past.

bwoshea:
> @WillHicks96 The PR itself seems fine to me, but (1) the PR needs to be merged with the tip of the main branch, and (2) the tests are failing due to a Grackle issue. Can you address these? I'm also going to check on Slack to see if people know how to fix the Grackle problem - I vaguely recall this happening in the recent past.

Following up on this - this was fixed in PR#306, so if you merge with the tip of `main` and fix the merge conflicts it should solve all of the problems.

PR #304, opened 2023-04-12
rescaling color fields when enforcing color fields

WillHicks96:
This is a very small PR that rescales color fields when the density floor is enforced in `enzo_SolveHydroEquations`. When the density floor is enforced, mass is effectively being added to the system. This change makes sure that the various species concentrations are kept the same when this happens.

### DIFF
--- a/src/Enzo/enzo_SolveHydroEquations.cpp
+++ b/src/Enzo/enzo_SolveHydroEquations.cpp
@@ -144,7 +144,17 @@ int EnzoBlock::SolveHydroEquations
 
   if (density_floor > 0.0) {
     for (int i=0; i<mx*my*mz; i++) {
-      density[i] = std::max(density[i], density_floor);
+      double d_pre = density[i];
+      density[i] = std::max(d_pre, density_floor);
+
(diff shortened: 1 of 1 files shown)


mabruzzo (approved):
Looks great!

jwise77:
Unless I'm missing something, you're rescaling all the cells, regardless of whether the density floor is breached. I would actually not use `std::max` and just have an if-statement comparing the density and the floor. Then inside of that block, alter the color fields.

jwise77 (approved):

mabruzzo:
@WillHicks96 - I think this is ready to be merged after resolving the conflicts. Would you like me to take a stab at that? (Getting this merged in would be helpful for some stuff that I would like to get done).

WillHicks96:
> @WillHicks96 - I think this is ready to be merged after resolving the conflicts. Would you like me to take a stab at that? (Getting this merged in would be helpful for some stuff that I would like to get done).

@mabruzzo Just took care of it -- was just a quick update. Should be good to go now!

PR #306, opened 2023-05-10
Update grackle branch name from master to main (in CI)

mabruzzo:


jwise77 (approved):

brittonsmith (approved):

mabruzzo:
The second commit introduces a hack to account for the fact that the type of the ``grackle_data_file`` field of Grackle's ``chemistry_data`` struct was changed from ``char*`` to ``const char*``.

This hack will no longer be necessary after we merge PR  #290

PR #313, opened 2023-05-25
Texascale bug fixes

jobordner:
Several issues were found attempting cosmology runs on 212992 cores of Frontera at TACC in April. This PR fixes several bugs, plugs several memory leaks, reduces checkpoint file sizes (by > 3 X when 16^3 blocks are used), adds some missing documentation, and refactors some entry method message classes.

  - **bug**: fixed hanging on restart due to state information not being initialized correctly on all processes
  - **bug**: fixed integer overflow in EnzoMethodBalance when processor count > 65536
  - **bug**: plugged memory leaks due to using CkFreeMsg instead of delete; affects restarted jobs in particular
  - **clean**: simplified size() save() load() methods in various entry method message classes
  - **perf**: added `Method:check:include_ghosts` parameter to allow omitting ghost zones in checkpoint files
  - **doc**: added documentation for `"check"` method parameters (including new `include_ghosts` parameter)


doc/source/param/method_check.incl
```diff
@@ -0,0 +1,81 @@
...
+dumps, though currently all data (all fields and all particles on all
+blocks) are written to disk.  This method should typically be called
+with a schedule.
+
+**Warning 1: Currently requires the** ``"order_morton"`` **method to be called beforehand, with a matching schedule.**
+
+**Warning 2: If a file or directory corresponding to a requested checkpoint
+dump already exists, no data will be written.** This is to avoid over-writing
+checkpoint files, but may not necessarily be what is expected. This behavior
+may be revised in the future.
```
mabruzzo, line 12: One alternative would be to try appending some characters to the file/directory name in order write something...

But that may be too complicated and it's definitely beyond the scope of this PR

mabruzzo (approved):
This looks great! I have 2 small suggestions, but otherwise I'm happy for you to merge whenever you are ready.

doc/cached_mtimes.json
```diff
@@ -0,0 +1 @@
+{"src": 1684971048, "dox": 1684971572}
```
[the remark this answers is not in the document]
jobordner: yes, not sure where this came from

doc/source/param/method_check.incl
jobordner: Agreed

jobordner:
@mabruzzo I realize I've made a number of additional commits after your review approval. I've re-requested a review should you choose to accept, and if I make any further changes I'll do that in a different branch.

mabruzzo:
Sounds great!

mabruzzo (approved):
This generally looks good to me. I didn't dig into the newer parts of the changes as deeply as my initial review (I was finding it a little difficult with all of the other changes), but I'm content to move forward.

I like the fact that we now avoid writing a checkpoint immediately after a restart.

WillHicks96 (approved):

PR #318, opened 2023-06-12
Improve ergonomics of pytest machinery

mabruzzo:
This PR introduces a number of small changes that should make the pytest machinery easier to use, particularly when you are switching between multiple branches.

These changes include:
1. Introducing the ability to configure machinery by passing command-line arguments in addition to setting environment variables. Below I list new command-line arguments with their env variable counterparts (when both counterparts are set, the command-line flags have priority):
    - `--charm` and `CHARM_PATH`
    - `--grackle-input-data-dir` and `GRACKLE_INPUT_DATA_DIR`
    - `--local-dir` and `TEST_RESULTS_DIR` (the command line flag borrows the name that `yt` uses)
    - `--answer-store` and `GENERATE_TEST_RESULTS` (the command line flag borrows the name that `yt` uses)
2. I introduced the new flag called `--build-dir` which you can use to specify the directory where the target Enzo-E binary was built. If `--charm` and `CHARM_PATH` are not specified, then the test program will infer the appropriate location of the `charmrun` executable based on the contents of `CMakeCache.txt`.
3. I added the ability for the testing infrastructure to directly query what the compiled-precision is (obviating the need for the `USE_DOUBLE` environment variable), and querying whether the executable was built with Grackle (tests no longer fail problem if `GRACKLE_INPUT_DATA_DIR` is set and enzo-e was compiled without Grackle - those tests are now skipped).

As an aside, I completely removed the `ENZO_PATH` variable as it's no longer necessary (and I'm pretty sure that I'm the only person to use it). Everything else should be fully backwards-compatible.

-----

With these changes, it's much easier to invoke the pytest machinery while moving between different branches that are built in different build directories (and may depend on different versions of charm++).

For example, if I have the build-directories
- `build-intel-impi`
- `build-gcc-openmpi`

I can invoke `pytest` from the root directory with:
- ``pytest test/answer_test --build-dir ./build-intel-impi --local-dir ~/enzoe-gs-intel``
- ``pytest test/answer_test --build-dir ./build-gcc-openmpi --local-dir ~/enzoe-gs-gcc``

While this is verbose, I would have previously needed to be sure to update the `ENZO_PATH`, `CHARM_PATH`, and `TEST_RESULTS_DIR` environment variables. I might also need to adjust `GRACKLE_INPUT_DATA_DIR` or `USE_DOUBLE` if one of these builds was compiled with/without grackle or with a different precision. 

Additional improvements could potentially be made. For example, it may be sensible for `pytest` to assume that it is invoked from a build-directory when it is invoked without `--build-dir` (unless it is invoked from the root directory). But, that's a topic for another day.

----

In the course of this PR, I refactored the functionality for executing an `enzo-e` simulation into a class called `EnzoEDriver` and moved it into a new `test_utils` sub-directory.

- Currently there are a bunch of wrappers for executing `enzo-e` floating around the repository. The idea was to provide a single centralized implementation that other python code can call into. While most of these other wrappers occur in test code that should eventually be integrated into the pytest, I expect it may still be useful to have this centralized wrapper.

- In any case, it was important to factor out the enzo-e wrapper from the rest of the answer-testing machinery. This will be important for new checkpoint-restart tests that I'm adding to the pytest machinery, which are nearly ready for review.

- Along the way, I also added the ability to allow the enzo-e wrapper to easily modify the parameters used in the simulation. This was inspired by what athena++ does and greatly simplifies the code for forthcoming checkpoint-restart tests and convergence tests. (Note: modifying the parameters currently happens outside of the enzo-e binary, by creating a new parameter file)

**EDIT:** Reviewing this PR is important because it is a dependency of PR #324, which should be a high priority

### DIFF
--- a/doc/source/tests/pytest_framework.rst
+++ b/doc/source/tests/pytest_framework.rst
@@ -9,6 +9,9 @@ simulation is run with two versions of ``enzo-e`` and their results are compared
 This is useful for testing problems with no analytical solution or generally
 verifying that results from commonly run simulations don't drift.
 
+It is also useful for testing property of simulation runs beyond the exit-time/cycle of the simulation.
+(While such tests do exist in the ctest-framework, they often involve more boiler-plate code).
+
 `pytest <https://docs.pytest.org/>`__ is a Python-based framework for detecting
 and running a series of tests within a source code repository. When running
 ``pytest``, the user can provide a directory in which ``pytest`` will look for
@@ -41,6 +44,8 @@ other useful answer testing functionality are located in the source in
 `test/answer_tests/answer_testing.py`. All answer tests are located in the
 other files within the `test/answer_tests` directory.
 
+Some other functionallity, that may be reused in other unrelated scripts provided in the Enzo-E repository, are provided in the ``test_utils`` subdirectory.
+
 Running the Answer Test Suite
 -----------------------------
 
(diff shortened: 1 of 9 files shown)


doc/source/tests/pytest_framework.rst
```diff
@@ -60,16 +65,49 @@ To generate test answers, use the highest numbered gold standard tag.
...
+
+   * - flag
+     - env var
+     - description
+   * - ``--build-dir``
+     - N/A
+     - points to the build-directory where the target enzo-e binary was built (that binary has the path: BUILD_DIR/bin/enzo-e).
+       The path to the charmrun launcher will be inferred from the `BUILD_DIR/CMakeCache.txt` file, but can be overwritten by the ``--charm`` flag or the ``CHARM_PATH`` environment variable.
+       This precedence was chosen in case a user causes a change to relevant cached build-variables, but have not rebuilt Enzo-E (i.e. `CMakeCache.txt` may not be valid for the binary).
+       When this flag isn't specified, the test infrastructure searches for the enzo-e binary at ENZOE_ROOT/build/bin/enzo-e, but does'nt try to infer charmrun's location from `CMakeCache.txt`.
```
gregbryan, line 94:
```suggestion
       When this flag isn't specified, the test infrastructure searches for the enzo-e binary at ENZOE_ROOT/build/bin/enzo-e, but doesn't try to infer charmrun's location from `CMakeCache.txt`.
```

doc/source/tests/pytest_framework.rst
```diff
@@ -86,18 +124,12 @@ First, check out the highest numbered gold standard tag and compile ``enzo-e``.
    $ git checkout gold-standard-1
```
gregbryan, line 124: I know this is not changed in this PR, but aren't we on gold-standard-004?

test/answer_tests/conftest.py
```diff
@@ -0,0 +1,184 @@
...
+
+    # step 3a: determine the path to the enzo-e binary
+    if build_dir is not None:
+        enzoe_path = os.path.join(build_dir, 'bin/enzo-e')
+        if not os.path.isfile(enzoe_path):
+            raise RuntimeError(f"enzo-e isn't at {enzoe_path}.")
+    else:
+        enzoe_path = os.path.join(_root_dir, 'build/bin/enzo-e')
+        if not os.path.isfile(enzoe_path):
+            raise RuntimeError(f"enzo-e isn't at {enzoe_path}. If the it is "
```
gregbryan, line 130:
```suggestion
            raise RuntimeError(f"enzo-e isn't at {enzoe_path}. If it is "
```

gregbryan (approved):
Looks good to me!

brittonsmith (approved):
Yeah, this looks good. Just a couple typos I caught, but happy for this to be merged.

doc/source/tests/pytest_framework.rst
```diff
@@ -9,6 +9,9 @@ simulation is run with two versions of ``enzo-e`` and their results are compared
 This is useful for testing problems with no analytical solution or generally
 verifying that results from commonly run simulations don't drift.
 
+It is also useful for testing property of simulation runs beyond the exit-time/cycle of the simulation.
```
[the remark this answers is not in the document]
mabruzzo: Thanks for catching this! I'm not sure what I was saying here either... (I was probably referencing the fact that most tests in the `ctest` framework just check the final simulation time/cycle)

I replaced the whole paragraph, which was originally written as 

```rst
It is also useful for testing property of simulation runs beyond the
exit-time/cycle of the simulation.  (While such tests do exist in the
ctest-framework, they often involve more boiler-plate code).
```

with the following snippet:

```rst
It is also useful in for testing problems that do have analytic
solutions (the answer test might quantify how close a simulation
result is to the analytic expected solution).  While such tests do
exist in the ctest-framework, they often involve more boiler-plate
code.
```

doc/source/tests/pytest_framework.rst
mabruzzo: You are correct. I have updated this accordingly (and I've added a note commenting that readers will need to update the number in the future)

mabruzzo:
@brittonsmith and @gregbryan -- let me know if my changes addressed your concerns

gregbryan (approved):
Looks good to me!

brittonsmith (approved):
All good here.

PR #319, opened 2023-06-28
Addressed a minor bug that was causing clang to encounter compilation errors.

mabruzzo:


gregbryan (approved):

jobordner:
The gravity_vlct_stable_Jeans_wave test fails for me with clang, but that's probably unrelated.

jobordner (approved):

mabruzzo:
Thanks for catching that I'll open an issue for it. The failure of that test is really weird! The only thing I could possibly imagine  (since there shouldn't be **any** floating-point optimizations) is that default round-off behavior for floating-point values could hypothetically change.

PR #320, opened 2023-06-28
Compile Time Improvement: Isolate boost::filesystem as a depenendency of cello.cpp

mabruzzo:
Out of curiosity, I was using clang's -ftime-trace flag to test profile where most of the compile time was being spent. Unsurprisingly, a lot of the time (maybe 30% - 50% per source file is spent reading in header files). A lot of this time is spent reading the ``<charm++>`` and C++ standard library headers. I hope to look at that in the future.

Somewhat surprisingly, every source file was spending ~0.4 seconds just reading <boost/filesystem> (this may be a slight underestimate). Given that we have over 220 source files, that's about 80 seconds per build (in serial). For context, a serial build on my machine probably takes ~15 minutes (maybe more). This was not very rigorous, but it seemed like an easy win to slightly improve build times by a few percent.


gregbryan (approved):
Looks good (but will need to be modified since png was also removed)

jwise77:
@mabruzzo Can you resolve the conflicts? Then we can merge since I've marked this as a 1 reviewer PR.

bwoshea:
This PR now passes all of the tests and is up-to-date with the tip of `main`.  Time to merge!

PR #321, opened 2023-06-28
Speed up CI: Modify CMake files to skip unit-test binaries when `BUILD_TESTING` option is `OFF`

mabruzzo:
`BUILD_TESTING` is a commonly occuring option across many different CMake projects. In fact, it is automatically defined by CTest, and was already used to determine whether parts of CTest machinery should be defined. The option is automatica a default value on `ON` (users need to explicitly opt out of building it).

This PR made 3 very small changes:
1. When `BUILD_TESTING` is disabled, we no longer build the unit-test binaries.
2. Documentation about this option has been added.
3. The CircleCI file has been updated to set `BUILD_TESTING=OFF` when building Enzo-E for the purposes of running answer testing

On my personal machine, using the clang compiler and 4 cores, the build step takes 3 minutes 45 seconds when `BUILD_TESTING=ON` and 3 minutes 7 seconds when  `BUILD_TESTING=OFF`.

- This time savings arises because we are skipping the building and linking of ~40 c++ files.
- Since the answer-testing CircleCI job needs to build Enzo-E twice per run (once to generate test answers and a second time to run the tests) and `BUILD_TESTING` has no impact on answer testing, this should save a bunch of time. (Note: we won't see any time-savings in the first build until the Gold standard gets updated again)

I think these are all very sensible changes, since the user builds these tests by default.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -112,6 +112,10 @@ commands:
       usegrackle:
         type: boolean
         default: true
+      setupCTests:
+        type: boolean
+        default: true
+        description: whether to build CTest machinery (has no effect on pytest framework)
     steps:
       - run:
           name: "Checkout target tag from enzo-e repository."
@@ -128,9 +132,6 @@ commands:
             source $BASH_ENV
             source $HOME/venv/bin/activate
 
-            # convert boolean parameter to an env var storing 0 or 1
-            USE_DOUBLE=$(( 0 <<# parameters.usedouble >> + 1 <</ parameters.usedouble >> ))
-            USE_GRACKLE=$(( 0 <<# parameters.usegrackle >> + 1 <</ parameters.usegrackle >> ))
             # remove build directory in case we've already compiled before
             rm -rf build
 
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -128,9 +132,6 @@ commands:
             source $BASH_ENV
             source $HOME/venv/bin/activate
 
-            # convert boolean parameter to an env var storing 0 or 1
-            USE_DOUBLE=$(( 0 <<# parameters.usedouble >> + 1 <</ parameters.usedouble >> ))
```
gregbryan, line 132: I don't totally follow why this we no longer need to set USE_DOUBLE... (being lazy -- I haven't looked carefully but this does seem to be used in the answer_tests).

mabruzzo: It turns out we weren't actually using this variable for anything. We are directly using `parameters.usedouble` in the call to `cmake` (see line 141 of this file)

gregbryan: But doesn't this set an environment variable which is (potentially) used on line 14 of test/answer_tests/test_grackle.py?

mabruzzo: Right, that's true! But there's a subtlety here. 

Each job in our CI is currently designed so that each job is composed of a sequence of reusable commands:
- Environment variables are not carried over from one command to the next. 
- Tangentially, this is why lots of commands begin with `source $BASH_ENV` - it restores a restore a set of environment variables from a file that we wrote to disk in the "set-env" command
-  The highlighted snippet was part of a command that builds enzo-e
- the `pytest` framework is invoked in a separate command. In that command ([it starts here](https://github.com/enzo-project/enzo-e/blob/642d7d7e58ba5f9dd78c107ff2d6f861621aa8ce/.circleci/config.yml#L212)), we set up the necessary environment variables before invoking `pytest`

gregbryan: Ah, OK! Thanks, that explain a number of things.

brittonsmith (approved):

PR #322, opened 2023-06-29
Introduce Precompiled Headers to Speed up Compilation 

mabruzzo:
This introduces the ability to use precompiled headers to speed up the compile times. I had done some light profiling of compile-times with clang's `-ftime-trace` command in order to see where time was being spent during compilation time. 

- As I suspected, we spend a lot of time reading in headers (since every source file reads more headers than necessary). But, the details really surprised me.
- I had been concerned that my `View` template class was increasing compile times by a lot, but it took a negligible amount of time.
- instead a bunch of time is spent reading headers like `<random>`, `<vector>`, `<charm++.h>`, and `<boost/filesystem.h>`. (That last case is addressed separately in PR #320).

To mitigate this to some extent, I started to make use of precompiled headers. This is a compiler-specific feature that cmake provides a nice abstraction for. 
- Essentially a compiler will preparse a set of header-files and store the binary representation known as a precompiled header. 
- There is still a compile-time penalty for unnecessarily including headers (e.g. the compiler still spends a bunch of time instantiating lots of charm++ template specializations in every source file), but this is definitely helpful.
- There is a potential drawback - if we aren't careful, source files that make use of a precompiled header know about all symbols declared in the preparsed header (even if that that header would not be included in the source file).

For that last reason, our usage of precompiled headers is fairly conservative. We just precompile `<random>`, `<string>`, `<vector>`, `<charm++.h>`, and `"pup_stl.h"`. These are all included in `cello.hpp`. And we try to only use the precompiled header in source files that include the `cello.hpp` file. 
- in the worst case scenario, where a source file doesn't include `cello.hpp`, it will just know about some functions/classes in the `std` namespace and from the charm++. This seemed better than also exposing symbols from within Enzo-E
- I also added a compile-test to circleci to confirm that Enzo-E will continue to compile when we disable these precompiled headers. Along the way I refactored some existing CircleCi tests, so that they were implemented in a slightly more concise manner (I confirmed that they still work).

**Some Numbers:** On my machine, with the clang compiler and 4 cores, this reduces compile-times from 3 min 45 seconds to 3 minutes 12 seconds (A ~15% speedup). 
- This doesn't include the speedup from PR #320 - that's maybe another 5-10 seconds faster on my machine in the same configuration.
- Keep in mind that unit-test binaries generally account for ~30 seconds of that time (see PR #321). Those don't get rebuilt very frequently during incremental rebuilds.

As an experiment, I tried using a precompiled header for enzo.hpp for all files in the Enzo-layer and that shaved off another 25 seconds from my build. But, due to the concerns cited earlier, I don't think it's a good idea to go that route (unless we come up with a more robust testing methodology).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,6 +23,10 @@ commands:
   install-dependencies:
     description: "Install dependencies."
     steps:
+      - restore_cache:
+          name: "Restore dependencies cache."
+          key: dependencies-v3
+
       - run:
           name: "Install dependencies."
           command: |
@@ -56,6 +60,12 @@ commands:
             cd $HOME/yt
             export MAX_BUILD_CORES=2
             pip install -e .
+
+      - save_cache:
+          name: "Save dependencies cache"
+          key: dependencies-v3
+          paths:
+            - ~/local
 
   install-grackle:
     description: "Install grackle."
@@ -254,18 +264,7 @@ jobs:
     steps:
       - checkout
       - set-env
-
-      - restore_cache:
-          name: "Restore dependencies cache."
-          key: dependencies-v3
-
-      - install-dependencies
-
-      - save_cache:
-          name: "Save dependencies cache"
-          key: dependencies-v3
-          paths:
-            - ~/local
+      - install-dependencies # (handles caching of dependencies too)
 
       - when:
           condition: << parameters.usegrackle >>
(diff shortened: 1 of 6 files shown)


gregbryan (approved):
Approved.

jwise77 (approved):

PR #323, opened 2023-06-30
Introduce a check for bugs in answer-testing code

mabruzzo:
We generally expect an answer-test to fail for one of 3 reasons:
  1. A change was made to Enzo-E or one of its dependencies that changed the actual answer.
  2. A new answer-test was introduced (and the older gold-standard doesn't produce an answer to test against)
  3. The answer-test itself was modified and the information checked by the answer test has changed.

If we expected the answer-tests to fail for one of the above reasons, then we generally want to update the gold-standard. However, there is a fourth reason the answer test may fail - if a change was introduced to the answer-testing code that broke an existing test.

This introduces a change to explicitly check that this last case isn't a problem. When the answer-test fails, we will now regenerate the answer tests using the currently checked out version of the code and we will confirm that the answer tests pass while using that version.

To confirm that this change works as intended, I have intentionally changed the gold-standard mentioned in the circleci file to an old version. I will revert it after I confirm the update to the circleci job has the intended impact.


gregbryan (approved):

jwise77 (approved):
Looks good, but we'll have to resolve some conflicts after #322 is merged.

jwise77:
> Looks good, but we'll have to resolve some conflicts after #322 is merged.

No conflicts after #322 was merged, so we're good to merge this one in!

PR #325, opened 2023-07-05
C++ version bump from C++11 to C++17

### LINKED ISSUE #317: Bumping Minimum C++ version to C++17
Since there's ongoing work to build Enzo-E with C++ compilers dependencies like Kokkos or libtorch, which both use C++ 17, I was wondering if there would be any opposition to bumping the version used within Enzo-E to C++17.

This would afford a few advantages:
- There are a handful of places in the codebase where [std::variant](https://en.cppreference.com/w/cpp/utility/variant) could be useful for significantly simplifying code (namely in the EOS functionality, but also in a few other places like `EnzoFieldAdaptor`, `ViewCollec`, or maybe even in the `Param` class.)
- A while back I determined that the internals of `ParticleData` technically invokes a bunch of undefined behavior (this is related to the way we cast pointers... I have notes on this somewhere). In practice, this behavior will probably never be a problem. As I recall, the "correct" solution has something to do with using [std::launder](https://en.cppreference.com/w/cpp/utility/launder)
- I think we can drop boost as a dependency (I believe all of the ``<filesystem>`` functions are now implemented as part of the standard library)
- There are some other (less important perks as well)

While charm++ still gets built with C++11, I'm skeptical that it would really be an issue. Do people have thoughts?

mabruzzo:
A number of advantages are highlighted in Issue #317 (C++17 adds a number of useful new features).

The main reasons NOT to do this would be compiler-support. At this point most major compilers support C++17. But since there is ongoing work to introduce dependencies on libraries that require C++17 (like Kokkos or libtorch), this is something of a moot point. Furthermore, there are [discussions of bumping the minimum required compiler version for charm++ to C++17](https://github.com/UIUC-PPL/charm/discussions/3684).

This resolves #317.

This PR includes some minor refactoring of `ViewCollec` to make use of `std::variant`. While similar refactoring is now possible in other areas of the code, I wanted to limit the size of this PR. This fix is somewhat important for robustly addressing some compiler-portability issues that keep cropping up in this area of the code (see PR #309 and PR #319).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,8 +19,8 @@ endif()
 #Define project and languages
 project(Enzo-E VERSION 1.0.0 LANGUAGES C CXX Fortran)
 
-# We need C++11
-set(CMAKE_CXX_STANDARD 11)
+# We need C++17
+set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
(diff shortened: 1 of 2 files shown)


gregbryan (approved):

bwoshea (approved):
This looks good to me.  I think it's reasonable to get rid of the Boost dependency (in general I'm a huge fan of eliminating any dependencies), but that also doesn't need to hold up this PR.  Also, Boost is pretty readily available.

PR #326, opened 2023-07-07
Convert `EnzoEFltArrayMap` to `ViewMap<T>` (& resolve issue #172)

### LINKED ISSUE #172: Unsound `const` semantics in `EnzoEFltArrayMap`

mabruzzo:
This PR migrates the contents of `EnzoEFltArrayMap`, which is a class encapsulating a map of `CelloView<enzo_float,3>` instances, to a new class template called `ViewMap<T>`. For backwards compatibility, `EnzoEFltArrayMap` is now a type-alias of  `ViewMap<T>`.

## Overview

In general, I think it makes a lot of sense for defining ``ViewMap`` alongside ``CelloView`` and it's something I've talked about doing for a long time. A major motivation for doing this is to resolve #172, which details some soundness issues with `EnzoEFltArrayMap`. The soundness issue and solution are described down below in the following spoiler tag.

<details> 
  <summary>Soundness problem & Solution </summary>

- Historically, we tried to define ``EnzoEFltArrayMap`` such that a ``const EnzoEFltArrayMap`` represented read-only data (we defined a few extra methods to help facilitate this). This can be useful when declaring function arguments.
- However, it later came to my attention that there was an easy way to mutate the data contained within a ``const EnzoEFltArrayMap`` that is passed to a function. The only way to fix is to disallow the creation of copies of ``EnzoEFltArrayMap`` (and to disable *move* semantics as well).
- To fix this I adopted a solution that is very similar to what we do in `CelloView` for this exact scenario. `ViewMap<T>` now represents the mutable datastructure while `ViewMap<const T>` represents the immutable data structure. I have defined a special constructor to facilitate implicit casts from `ViewMap<T>` to `ViewMap<const T>` that are about as seamless as the casts from `float*` to `const float*`.
- I have removed some methods that are now made unnecessary after making this change.
- I have added a section to the website-documentation discussing this behavior.

</details>

When reviewing this PR, I suggest that you step through the commits step-by-step (rather than just looking at the final files). I essentially completed the following steps:

1. I defined `ViewMap<T>` which is almost a perfect one-to-one copy of `EnzoEFltArrayMap` (a few minor changes were needed to make it work as a class template).
2. I wrote a number of unit tests for `ViewMap<T>` (to help do this, I factored out some functionality from `Cello/test_ViewCollec.cpp` into a new header file called `Cello/test_ViewTestRoutines.hpp`). These tests mostly check existing functionality that we already know works (since we've been using `EnzoEFltArrayMap` for a while).
3. I removed `EnzoEFltArrayMap` from the codebase and made `EnzoEFltArrayMap` into an alias for `ViewMap<enzo_float>`. I did not replace `EnzoEFltArrayMap` throughout the codebase because that would touch a LOT of files (we can revisit that in the future). I also updated the website-documentation accordingly.
4. I now refactored `ViewMap` to actually address the soundness problem (and I updated the website-documentation accordingly)
5. I updated the internal comments and variables within `ViewMap` so that they now discuss the contained views rather than the contained arrays (which is consistent with our previous renaming of `CelloArray` -> `CelloView`). I did NOT touch the names of methods because that would touch a LOT of additional files (we can also revisit this in the future).

## Question for reviewers?
Do you think it's worth renaming `ViewMap` to `View3DMap` to make it clear that it only holds 3D views (i.e. `CelloView<T,3>`)?


doc/source/devel/array.rst
```diff
@@ -853,75 +861,133 @@ or to replace the ``CelloView`` associated with a given
...
+equivalent to ``scratch_view_map[key_l[i]]`` and
+``other_view_map[other_key_l[i]]``, in the context of the preceding
+snippets. Note that we don't support passing an integer value to
+``ViewMap::at``.
 
 Copy and ``const`` Semantics
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Making a copy of an ``EnzoEFltArrayMap`` instance (e.g. with a copy
+Making a copy of an ``ViewMap`` instance (e.g. with a copy
```
brenjohn, line 876:
```suggestion
Making a copy of a ``ViewMap`` instance (e.g. with a copy
```

doc/source/devel/array.rst
```diff
@@ -853,75 +861,133 @@ or to replace the ``CelloView`` associated with a given
...
-In comparison, element-access of a ``const EnzoEFltArrayMap`` yields a
-``CelloView<const enzo_float,3>`` which prevents direct modification of
-array elements.
+A ``const ViewMap`` holds very little meaning. If you declare and
+assign a variable, ``var``, as
+``const ViewMap<double> var = /*... */;``,
+then the fact that you used ``const ViewMap`` effectively tells the
+compiler to forbid you from performing ``var = other_view_map;`` in a
+subsequent statement.  You are still free to mutate the elements held
+of any contained CelloView.
```
brenjohn, line 889:
```suggestion
by any contained CelloView.
```

doc/source/devel/array.rst
```diff
@@ -853,75 +861,133 @@ or to replace the ``CelloView`` associated with a given
...
+
+If you want to denote that a ``ViewMap`` is read-only, you should
+declare declare the type as ``ViewMap<const double>`` instead. Note
```
brenjohn, line 892:
```suggestion
declare the type as ``ViewMap<const double>`` instead. Note
```

doc/source/devel/array.rst
```diff
@@ -853,75 +861,133 @@ or to replace the ``CelloView`` associated with a given
...
+that a ``ViewMap<float>`` can be implicitly converted to a
+``ViewMap<const float, N>`` (e.g. you can pass the former to a
```
brenjohn, line 894:
```suggestion
``ViewMap<const float>`` (e.g. you can pass the former to a
```

doc/source/devel/array.rst
```diff
@@ -853,75 +861,133 @@ or to replace the ``CelloView`` associated with a given
...
+    CelloView<enzo_float,3> calculate_pressure(ViewMap<const enzo_float> arg,
+                                               double density_floor);
+
+This function signature makes a very clear promise to other developers
+that ``arg`` is not mutated inside of the function. In fact the
+compiler will not compile any implementation of the function where the
+contents of arg are mutated (this can be a very useful reminder to any
+future developer that alters the implementation of
+``calculate_pressure`` - they are of course able to change the
+function signature, but it's a a good reminder to make sure that they
```
brenjohn, line 917:
```suggestion
function signature, but it's a good reminder to make sure that they
```

doc/source/devel/array.rst
```diff
@@ -853,75 +861,133 @@ or to replace the ``CelloView`` associated with a given
...
-     ``CelloView`` values in a single ``CelloView<enzo_float, 4>``
-     instance. Access of individual ``CelloView`` values is
-     accomplished with the overload of the
-     ``CelloView<T,D>::subarray`` method. This approach is used when
-     you construct an ``EnzoEFltArrayMap`` that allocates memory for
-     the contained ``CelloView``\s.
+     ``CelloView`` values in a single ``CelloView<T, 4>`` instance.
+     Access of individual ``CelloView`` values is accomplished with
+     the overload of the ``CelloView<T,D>::subarray`` method. This
+     approach is used when you construct an ``ViewMap`` that
```
brenjohn, line 989:
```suggestion
     approach is used when you construct a ``ViewMap`` that
```

src/Cello/view_ViewMap.hpp
```diff
@@ -0,0 +1,395 @@
...
+  /// key-order matches expectations.
+  ///
+  /// @param ref The list of reference keys
+  /// @param raise_err When true, this will raise a fatal error if the order
+  ///    doesn't match
+  /// @param allow_smaller_ref When `true`, the key-order is only compared for
+  ///    first `ref.size()`. When `false` (the default), the map must have the
+  ///    same number of keys as `ref`.
+  bool validate_key_order(const std::vector<std::string> &ref,
+			  bool raise_err, bool allow_smaller_ref = false)
```
brenjohn, line 179: Maybe it makes sense to have a default value for `raise_err` here also?

src/Cello/test_ViewMap.cpp
```diff
@@ -0,0 +1,765 @@
...
+    // getting the signatures on the template functions correct so that they
+    // can be used to compare the template functions is challenging! So we just
+    // do it manually!
+    bool all_equal = true;
+    for (const std::string& key: src_keys) {
+      CelloView<const int,3> dest_view = dest_view_map.at(key);
+      CelloView<const int,3> src_view = src_view_map.at(key);
+
+      const int mz = dest_view.shape(0);
+      const int my = src_view.shape(1);
```
brenjohn, line 685:
```suggestion
      const int my = dest_view.shape(1);
```
I guess this really doesn't make a difference but src_view is a little unexpected when reading this.

brenjohn (approved):
I think this looks great as it is. I only had very minor comments/suggestions.

brenjohn:
As for renaming `ViewMap` to `View3DMap`, I think that makes sense but there was a suggestion in the documentation to generalize the implementation which would make a 2D version redundant. Maybe it makes sense to add a note in the documentation that this only works for 3D views at the moment and then we can look at generalizing to 2D views when someone needs them?

src/Cello/test_ViewTestRoutines.hpp
```diff
@@ -0,0 +1,183 @@
+// See LICENSE_CELLO file for license and copyright information
+
+/// @file     test_ViewTestRoutines.cpp
+/// @author   Matthew Abruzzo (matthewabruzzo@gmail.com)
+/// @date     2023-07-03
+/// @brief    Common routines used for testing CelloView and related machinery
+
+#ifndef TEST_VIEW_TEST_ROUNTINES_HPP
```
bwoshea, line 8: I'm not entirely sure what a "ROUNTINE" is... I think that this is a typo (same note for the line immediately below)

bwoshea (approved):
Outside of one minor comment I don't have much to add beyond what @brenjohn had to say.  It looks good to me!

bwoshea:
@mabruzzo it looks like there are some minor suggestions that should be very easy to resolve, plus merge conflicts that need to be fixed.  If you can do those we can merge this PR!

src/Cello/view_ViewMap.hpp
mabruzzo: added

src/Cello/test_ViewMap.cpp
mabruzzo: Good point! fixed

src/Cello/test_ViewTestRoutines.hpp
mabruzzo: Thanks for catching that!

mabruzzo:
I know the answer-tests started failing before for commit [0d99309](https://github.com/enzo-project/enzo-e/pull/326/commits/0d9930996d4973d12996d370820bbd5371468e47).

But, I think that may be related to a grackle-change and the fact that we don't pin our dependencies entirely correctly. A little while back, we had to update the gold-standard to deal with this issue. I think that commit [0d99309](https://github.com/enzo-project/enzo-e/pull/326/commits/0d9930996d4973d12996d370820bbd5371468e47) didn't include those changes.

If I'm right about this, then merging in the main branch just now should cause the tests to pass.

PR #327, opened 2023-07-08
Fix Some Compiler Warnings in Cello Layer

mabruzzo:
This PR fixes some compiler warnings in the Cello-Layer. Between this PR and PR #298 just about all Compiler-Warnings are addressed in the Cello Layer. 

Along the way, one of the warnings led me to catch a minor bug with an error-check to the active-zone's size. While I was there I improved the error message formatting (so that when we print the shape of the active zone, we consider the simulation's dimensionality).

The remaining warnings in the Cello Layer errors have something to do with `strncpy`. This needs slightly more attention & can be handled in the future (or suppressed). There are a bunch of warnings that still need to be addressed in the Enzo-Layer (but that's something for the future).


src/Cello/mesh_Adapt.cpp
```diff
@@ -396,7 +395,7 @@ void Adapt::print(std::string message, const Block * block, FILE * fp) const
                  ia3[1],it3[1],
                  ia3[2],it3[2]);
       }
-      const int l = 1 << (max_level_ - level);
+      //const int l = 1 << (max_level_ - info.index_.level());
```
gregbryan, line 398: This is commented out -- should it stay for some reason?

gregbryan (approved):
Looks good!

src/Cello/mesh_Adapt.cpp
mabruzzo: I commented this out because it was an unused variable. I'm not sure if there was any reason to keep it... If you think I should delete it, I can do that

src/Cello/problem_MethodOutput.cpp
```diff
@@ -396,7 +399,7 @@ void MethodOutput::compute_done (Block * block)
 void Block::r_method_output_done(CkReductionMsg *msg)
 {
   delete msg;
-  MethodOutput * method = static_cast<MethodOutput*> (this->method());
+  //MethodOutput * method = static_cast<MethodOutput*> (this->method());
```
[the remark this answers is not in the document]
mabruzzo: I commented this out because it was an unused variable. I'm not sure if there was any reason to keep it... If you think I should delete it, I can do that

src/Cello/mesh_Adapt.cpp
gregbryan:
```suggestion
```
Let's rip off that band-aid...

src/Cello/problem_MethodOutput.cpp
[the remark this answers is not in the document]
gregbryan:
```suggestion
```

PR #329, opened 2023-07-17
Delete contents of `src/Enzo/obsolete`

mabruzzo:
During the reorganization of `src/Enzo` I created `src/Enzo/obsolete`, and all contents of that directory had been recommended by  @jobordner. This PR is removing those file from the codebase (if we really need them later, we can always look at an older revision).

The main contents of this source directory were `EnzoInitialPm` and `EnzoMethodHydro` 
- The former was used in some parameter files. None of these parameter files were involved in any automated tests. Thus, I deleted all of the associated parameter files.

This directory also contained 2 source files that didn't really accomplish anything. Thus I deleted those too.


gregbryan (changes requested):
This looks good to me but just wanted to double-check with @jobordner that InitialPm is really obsolete.

mabruzzo:
@gregbryan, I just wanted to clarify there aren't any particular changes that you actually want from me, right?

gregbryan:
> @gregbryan, I just wanted to clarify there aren't any particular changes that you actually want from me, right?

No -- and, in fact, I don't think we need to wait. That's the great thing about version control -- we can always bring code back!

gregbryan (approved):

jobordner (changes requested):
Please restore EnzoInitialPm.cpp.  Otherwise looks ok.

In general, if you find untested code, creating an issue noting that there is untested code is preferable to deleting the code.

mabruzzo:
I think there was a miscommunication back when we reorganized everything -- essentially, I thought that everything in the `obsolete` directory could be removed.

I'm happy to restore it (and all the relevant input files)

@jobordner I have 2 questions:
1. Do you want me restore `input/Obsolete/pm3.incl` and  `input/Obsolete/collapse-pm-3d.in` (the other ~10 files are located in different directories)? 
2. Is the plan to eventually remove `EnzoInitialPm`? 
- If so, I'll leave it in `obsolete` (or maybe I'll move it to `src/Enzo/initial/obsolete`).
- If not, would it make more sense for me to rename the initializer class to something like `EnzoInitialPmCollapseTest` or `EnzoInitialPmTest`? (In that case, I would move it to `src/Enzo/tests` (it contains other similarly named classes)

jobordner (approved):

PR #332, opened 2023-07-19
Convert `Enzo/hydro-mhd` into a subcomponent (Build Restructuring Part 1)

mabruzzo:
This refactors `Enzo/hydro-mhd` into "subcomponents" that are compiled separately from the rest of the enzo-layer. This should improve compilation-times and the time it takes to perform a rebuild after a small change to a header.

This is part of a general restructuring that I'm applying to the whole enzo-layer. A separate PR has been proposed, #333, to update the website documentation to explain my thought-process. 

However, it is worth separating this PR from #333. If we ultimately decide NOT to apply this restructuring strategy to the rest of Enzo-Layer (in which case #333 will be rejected), I think it makes a lot of sense for the `Enzo/hydro-mhd` subdirectory.

I will also post a series of other PRs that apply similar changes to other parts of the Enzo-Layer. (Part of the reason that they have been separated is that they touch a bunch of header files)


bwoshea (approved):
This looks good to me - the changes all seem straightforward!

bwoshea:
@mabruzzo this is ready to merge once the branch is updated to include changes from main!

PR #337, opened 2023-07-21
Introduce `cello::is_initial_cycle`

mabruzzo:
## Overview
This function abstracts some logic for checking whether the simulation was executing the very first cycle after starting up. 

This logic previously appeared in the following places:

* 3 methods of the `Block` class
* 9 Physics-Methods from the Enzo-layer (mostly for checking invariants)
* within `EnzoMethodCheck` to avoid writing a new checkpoint in the same cycle as a restart.

This should make it a lot easier to edit this logic in the future

## Reviewer Notes
This builds directly atop the changes introduced in #313. Only the last commit in this PR is different from that PR. I can reset-the file differences in this PR after #313 is merged


src/Cello/cello.cpp
```diff
@@ -565,4 +565,37 @@ namespace cello {
...
+  //----------------------------------------------------------------------
+
+  bool is_initial_cycle(int cycle, InitCycleKind kind) noexcept {
+    switch (kind) {
+    case InitCycleKind::fresh:
+      {
+        const Config * config = cello::config();
+        ASSERT("cello::is_initial_cycle(int, InitCycleKind)",
+               "cello::config() returned nullptr", config != nullptr);
+        return cycle == config->initial_cycle;
```
brenjohn, line 585: Just wondering if there's a preferred way of accessing the `initial_cycle` variable (via `cello::config` or `cello::simulation` as in line 596). I'm not sure if there's a subtle difference between these but if not, maybe it's better to use the `simulation` interface here rather than accessing the `config` member directly?

In that case, using `fresh_or_noncharm_restart` inplace of `fresh` when calling these functions could save a little bit of code duplication.

brenjohn (approved):

src/Cello/cello.cpp
mabruzzo: That's an excellent point!

It's been a long time since I wrote this. But, I vaguely recall that there is a difference in the 2 values based on whether or not we were doing a restart... (or I was concerned that the values might be different in the future). I'll look into that.

At the very least, I should document this with a comment.

WillHicks96: Just adding onto this. From what I can tell, `cello::simulation()->initial_cycle()` returns `cello::config()->initial_cycle`, so I don't think there is a difference between the two--though I haven't actually tested it to be sure.
You can see in `Simulation::initialize_simulation_()` where `cycle_initial_` is set to `config_->initial_cycle`.

WillHicks96 (approved):

PR #340, opened 2023-08-03
Introduce `EnzoPhysicsGravity`

### LINKED ISSUE #189: Inconsistent Gravitiational Constants
While implementing PR #186, I became aware of a potential issue.

At this time the following files in the `Enzo` layer make use of `cello::grav_constant`:
- `enzo_EnzoInitialIsolatedGalaxy.cpp`
- `enzo_EnzoMethodStarMaker.cpp`
- `enzo_EnzoMethodStarMakerStochasticSF.cpp`
- `enzo_EnzoMethodBackgroundAcceleration.cpp`
- `enzo_EnzoMethodStarMakerFIRE2.cpp`

I don't understand the context of this choice in any of these files, but I this is the wrong thing to do if you're interested in using the Gravitational constant value used by the simulation. The simulation instead uses the value specified by `Method:gravity:grav_const`.

EDIT: I had a followup conversation with @jobordner about this topic and he confirmed that this could be an issue and that some refactoring is probably needed here. If you want to use this feature in the immediate short-term, there are some workarounds for this.

mabruzzo:
### High-Level Summary

This PR defines ``EnzoPhysicsGravity``, whose purpose is to provide a single location where the gravitational constant is defined. This lets us assign an arbitrary value to the gravitational constant in non-cosmological simulations.

This PR resolves #189.

### Motivation

A common feature among hydro-codes is to allow the user to specify an arbitrary gravitational constant. This capability simplifies certain test problems. I wanted to introduce this capability to Enzo-E in non-cosmological simulations (this capability is not provided in cosmological simulations, since the time-code units are defined in terms of the Gravitational constant).

Prior to this PR, if a user ran a non-cosmological simulation with ``EnzoMethodGravity``, they were responsible for specifying the gravitational constant in code-units as the parameter ``"Method:gravity:grav_const"``.

* While the user could in principle specify an arbitrary value for the gravitational constant here, it would only be used within the "gravity" method.
* Just about all other areas of the code that had calculations involving the gravitational constant used the constant formerly called ``enzo_constants::grav_constant``.
* Thus, if the value passed to ``"Method:gravity:grav_const"`` was not equivalent to the standard value of the gravitational constant, the simulation might not be self-consistent.

### Overview

To make it possible to assign an arbitrary value to the gravitational constant, I defined the ``EnzoPhysicsGravity`` Physics object, whose sole purpose is to track the gravitational constant.

* The value is specified with the ``"Physics:gravity:grav_const_codeU"`` parameter. If a user forgets to specify this parameter in a non-cosmological sim, the gravitational constant is assumed to be the standard value. If the user tries to specify the parameter in a cosmological sim, Enzo-E will abort with an error.

* Care was taken to maintain backwards compatability: the physics object will also be configured when ``"Method:gravity:grav_const"`` is provided (but a warning is printed noting that the parameter is deprecated).

2 convenience functions were defined, ``enzo::grav_constant_codeU()`` & ``enzo::grav_constant_cgs()``, that retrieve the gravitational constant from the physics object (they also return the correct values in cosmological sims). Just about all occurences that referenced ``enzo_constants::grav_constant`` were replaced with calls to these 2 functions.

* ``enzo_constants::grav_constant`` was replaced in methods of the following classes: ``EnzoInitialIsolatedGalaxy``, ``EnzoInitialShuCollapse``, ``EnzoMethodBackgroundAcceleration``, ``EnzoMethodBondiHoyleAccretion``, ``EnzoMethodSinkMaker``, ``EnzoMethodStarMaker``, ``EnzoMethodStarMakerSTARSS``, ``EnzoMethodStarMakerStochasticSF``

* I also renamed ``enzo_constants::grav_constant`` so that it is now called ``enzo_constants::standard_grav_constant``, to convey that the variable tracks the standard choice for the gravitational constant (but may not be the actual value used in the simulation)

* Presently, only ``EnzoPhysicsCosmology`` and ``EnzoPhysicsGravity`` refer to ``enzo_constants::standard_grav_constant``.

I also modified 2 other areas of the code to make use of ``enzo::grav_constant_codeU()``:

* the initializer of the stable jeans wave test-problem from ``EnzoInitialInclinedWave``) (Previously we were doing something hacky involving ``enzo::config()``)

* the ``SetMinimumPressureSupport`` function. Previously, this function did not work properly for non-cosmological simulations. While I was there, I also fixed a bug in the way that this function was being used - previously it was used to apply a floor independent of refinment level, but in reality it should only be applied for the highest refinement levels.

Finally, I updated a bunch of documentation:

* I introduced user-facing documentation for ``EnzoPhysicsGravity``

* I added reference documentation for the ``"Physics:gravity:grav_const_codeU"`` parameter and modified the documentation for ``"Method:gravity:grav_const"``

* I updated the reference documentation for the minimum_pressure_support parameters (based on enzo-classic's docs)

* I added some developer documentation that tries to motivate the purpose of Physics classes and provide some examples of cases where they are particularly useful. This documentation is pretty rough, but it's definitely a start.

### Other Notes

While defining a whole Physics class to track a single variable seems like overkill, it turns out to be necessary.

However, the main alternative would be to continue to store the gravitational constant in the ``EnzoMethodGravity`` class. This is undesirable since there are certainly cases where somebody might want use a custom Gravitational Constant without using that particular method.

* This will become a common scenario once the new ``Method`` class implementing FMM is merged into the codebase.

* Furthermore, it's plausible that someone might want to use ``EnzoMethodBackgroundAcceleration`` without any self-gravity and adjust the gravitational constant in that case as well (e.g. in a test problem).

In the future, we could always rename/refactor ``EnzoPhysicsGravity`` so that it contains miscellaneous configuration-based data unrelated to gravity.


gregbryan (approved):
Looks good to me!

doc/source/devel/physics.rst
```diff
@@ -0,0 +1,97 @@
...
+However, because Enzo-E has :cpp:class:`!Method` classes that implement different hydro-solvers (that use the dual-energy formalism), we instead encode this information in a :cpp:class:`!Physics` class.
+
+There are also scenarios where some configuration information isn't really associated with any singular :cpp:class:`!Method`.
+For example the Equation-Of-State is important to a number of different methods.
+Another example includes the Gravitational Constant - this is important in self-gravity solvers and external-potential solvers, which are implemented in different :cpp:class:`!Method` classes.
+
+General Tips
+============
+
+The general advise is to implement a :cpp:class:`!Physics` class so that it is immutable (after construction the instance's state doesn't change).
```
clairekope, line 82:
```suggestion
The general advice is to implement a :cpp:class:`!Physics` class so that it is immutable (after construction the instance's state doesn't change).
```

doc/source/devel/physics.rst
```diff
@@ -0,0 +1,97 @@
...
+This makes the behavior of :cpp:class:`!Physics` classes much easier to reason about since a single PE only has one instance of a :cpp:class:`!Physics` given class, but is responsible for evolving one or more instances of :cpp:class:`EnzoBlock`.
```
clairekope, line 83: Not super familiar with the Enzo-e docs; if PE is common nomenclature this is probably fine
```suggestion
This makes the behavior of :cpp:class:`!Physics` classes much easier to reason about since a single processing element only has one instance of a :cpp:class:`!Physics` given class, but is responsible for evolving one or more instances of :cpp:class:`EnzoBlock`.
```

doc/source/devel/physics.rst
```diff
@@ -0,0 +1,97 @@
...
+
+Quirky Implementations
+======================
+
+:cpp:class:`!EnzoPhysicsCosmology` currently tracks some mutable state.
```
clairekope, line 88: Is it possible to list these? I assume the scale factor is one.

doc/source/param/method.incl
```diff
@@ -382,6 +362,22 @@ gravity
...
+   :Summary: :s:`Gravitational constant`
+   :Type:    :par:typefmt:`float`
+   :Default: :d:`none`
+   :Scope:     :z:`Enzo`
+
+   .. warning::
+
+      This parameter is deprecated and will be removed in the future.
+      The user should use :par:param:`Physics:gravity:grav_const_codeU`
+      instead. Thesen 2 parameters have identical behavior when using
```
clairekope, line 378:
```suggestion
      instead. These 2 parameters have identical behavior when using
```

doc/source/param/method.incl
```diff
@@ -382,6 +362,22 @@ gravity
...
+      the ``"gravity"`` method, but the new parameter influences other
+      gravity-related methods.
```
clairekope, line 380: Such as?

src/Enzo/gravity/EnzoPhysicsGravity.hpp
```diff
@@ -0,0 +1,152 @@
...
+  //      enzo_constants::standard_grav_const when converted to cgs
+  EnzoPhysicsGravity(double grav_const_code_units)
+    : Physics()
+  {
+    if (grav_const_code_units <= 0.0) {
+      grav_constant_codeU_ = -1.0;
+    } else {
+      ASSERT("EnzoPhysicsGravity::EnzoPhysicsGravity",
+             "users aren't allowed to specify the gravitational constant in "
+             "a cosmological simulation", enzo::cosmology() == nullptr);
```
clairekope, line 63: Should this be !=; that is, throw the error if the `enzo::cosmology()` pointer was initialized?

src/Enzo/hydro-mhd/SetMinimumSupport.cpp
```diff
@@ -48,11 +48,29 @@ int EnzoBlock::SetMinimumSupport(enzo_float &MinimumSupportEnergyCoefficient,
...
+    //   c_s * sqrt(pi / (G * rho)) >= sqrt(K) * CellWidth
+    //   c_s^2 *  pi / (G * rho) >= K * CellWidth^2
+    //   gamma * (gamma - 1) * eint * pi / (G * rho) >= K * CellWidth^2
+    //   eint >= G * K * CellWidth^2 * rho / (pi * gamma * (gamma - 1))
+    //   eint >= MinimumSupportEnergyCoefficient * rho
+    //
+    // I'm don't totally understand where the extra CosmoFactor comes in...
+    // that was here earlier
+
+    // TODO: check that CellWidth[0], CellWidth[1], and CellWidth[2] are
```
clairekope, line 68: Do you have plans to implement this TODO?

clairekope (changes requested):
Overall looks good; just some tweaks to documentation and two small code questions

src/Enzo/gravity/EnzoPhysicsGravity.hpp
mabruzzo: No this is correct as it's written. This is an assertion that `enzo::cosmology() == nullptr`. The error message is printed if that condition is violated.

I vaguely recall that I had things swapped around when I originally wrote this snippet (and I caught this mistake from a failing test)

doc/source/devel/physics.rst
mabruzzo: Good point

clairekope (approved):

clairekope:
@mabruzzo don't want to leave an unattended TODO floating in the code for too long but in the interest of pushing things along I'm approving :)

PR #341, opened 2023-08-07
Migrate Parameter Parsing: introducing `ParameterAccessor`

mabruzzo:
This PR is the first in a series of that alters the way in which classes (like ``Method`` or ``Physics``) are configured from values taken from the parameter file. 

## Plan to roll out these changes

We **need** to introduce these changes in discrete steps because (i) it would take a lot of work to create (and review) a PR migrating everything at once, and (ii) additional work needs to be done every time a new parameter gets added via the old approach (to convert it so that it uses the new approach).

With that in mind, this PR takes a large first step to begin overhauling parameter parsing. It starts by shifting the vast majority of ``Method`` config-parsing from ``Cello``'s ``Config`` object into the ``Method`` subclasses (the remaining stuff needs to be addressed after we finish the conversion in the Enzo-layer). It also overhauls the process of configuring some ``Method`` objects from the Enzo-layer

Additional PRs will follow that deal with migrating other ``Method`` classes in the Enzo-layer. The changes in many of these cases are slightly more involved/less straightforward. In the future I also plan to extend this to other parts of the codebase (like ``Initial`` subclasses or ``Physics`` subclasses).

## Motivation

Since I'm not rolling out this transition all at once, I have added quite a bit of website documentation for this new approach. This documentation explains the differences in the approaches and motivates why we are making this change:
- you can find this documentation in `doc/source/devel/param.rst`
- Alternatively, you can look at the preview website pages (built as part of this PR). In the sidebar, go to ``Developer Guide`` -> ``Parameter Parsing and Configuration``

With that said I will try to highlight some weaknesses (that emerged for historical reasons) of the old approach:

- Adding a new parameter is fairly laborious. To "properly" add a new parameter for a given ``Method`` subclass in the Cello layer (Enzo layer), one must store the parsed the value as an attribute of the global ``Config`` (``EnzoConfig``), forward it to the subclass, where the value is also stored as an attribute. This involves ~7 steps (see documentation).
- Because this is laborious, there is a strong temptation to cut corners. The common shortcut is to simply store the parsed value in the global ``Config`` (or ``EnzoConfig``) object and always access the value from there. This is equivalent to storing the parsed values as global variables (which is obviously undesirable for a number of reasons).
- the process of configuring multiple instances of a single ``Method`` subclass could be made cleaner

The change made to the parameter-parsing by this PR is fairly simple:

> The constructors (or static-factory-methods) of ``Method`` subclasses directly retrieve the parameter values from the parameter file and store the values as attributes. We completely no longer store attributes them as attributes within ``Config`` or ``EnzoConfig`` (in fact, all steps involving these classes are skipped).

This change reduces the number of steps needed to introduce a new parameter from 7 to 3, and strongly discourages the usage of global variables. This takes inspiration from Athena++

## The ``ParameterAccessor`` class

While one can imagine implementing our new approach by passing a reference to the ``Parameters`` class to the constructors of various ``Method`` classes, we instead choose to pass the newly defined ``ParameterAccessor`` class.

<details> 
  <summary>To understand ``ParameterAccessor``, I need to define a "parameter-path" (within the spoiler tag) </summary>

  ----
  Parameters in Enzo-E parameter files for are organized into hierarchical groups. This mirrors in which files are organized within a directory hierarchy. The following block shows a snippet from a hypothetical parameter file that configures ``EnzoMethodHeat``.

  ```
  Method{
    list = ["heat"];
    heat {
      alpha = 0.5;
    }
  }
  ```

  To unambiguously refer to a parameter in the file, we would refer to the parameters in this snippet of ``Method:list`` and ``Method:heat:alpha``. This shorthand is similar to file paths on a unix-like operating system (e.g. ``/home/matthew/my_file.txt``). For that reason, I will use term "parameter-path" to refer to these "full names of these parameters".

  ----
</details>

Essentially, the ``ParameterAccessor`` has an associated root-parameter path specified during its construction:
- All parameter paths passed to methods of this class are considered to be relative-paths to this root-location
- The documentation highlights how this class helps simplify code in the case where we want to construct multiple instances of a given ``Method`` subclass.
- Technically, the ``Parameters`` class supports a similar feature, but it's less obvious when code is relying upon that feature (when you see ``ParameterAccessor``, you know that the feature is being used).

## Notes for the reviewers

- Please carefully read the documentation. I found it fairly difficult to write. So any feedback would.
- Most of the changes outside of ``ParameterAccessor.[ch]pp`` just consist of moving parsing out of ``Config`` and ``EnzoConfig``
- I want your feedback about the following point (discussed in the following spoiler tag):

<details> 
  <summary> Addressing drawback of ``ParameterAccessor``</summary>

  ----
  The primary drawback of ``ParameterAccessor`` is that we no longer specify the full parameter-paths of every parameter. While this is fundamentally necessary for gracefully handling the case where we want to configure multiple instances of a given `Method` class, it's still less obvious for new users.

  My thought is: what it we extend the analogy between parameter-paths and file-paths? 
    - We could make a leading `":"` signify an absolute parameter-path (much in the same way that a leading `"/"` signifies an absolute file path). 
      - For example,  passing ``":Method:heat:alpha"`` to an instance method of ``ParameterAccessor`` would always refer to the same parameter regardless of the root-path currently associated with the instance of ``ParameterAccessor``
      - This could be used in cases where we don't expect more than 1 instance of the ``Method`` subclass to need to be configured in a given simulation (which is a lot of them).
    - We could also make a leading `".:"` signify that a parameter-path is explicitly a relative path (much in the same way that a leading `"./"` explicitly signifies  that a file-path is relative to the current directory).

  **Do you have any thoughts about this?**

  ----
</details>

## Future Plans:

- In the future, I would like intention to remove the code where ``Cello``/``Enzo-E`` automatically set the courant factor of each ``Method`` subclass. Instead, I want each ``Method`` subclass to parse and set their own courant factor, especially since the default courant value may differ from case to case (for example, the VL+CT solver requires a value at or below 0.5 while lots of other methods require a value at or below 1.0).
- Once we finish transitioning the ``Method`` objects, I want to start raising warnings (or optionally raising errors), when users specify parameters that aren't used for anything in a parameter file (usually this means there is some kind of typo - I've done this a lot).


mabruzzo:
@jobordner - I know a while back that you mentioned you had concerns about safe-parsing in SMP-mode. I can/should explicitly check that.

But, don't let this stop you from reviewing this PR. I'm very confident that there really shouldn't be any problems at all since I changed any of the underlying parsing code (or when the parsing is done)

src/Cello/problem_Problem.cpp
```diff
@@ -468,7 +468,15 @@ void Problem::initialize_method
 
   Method::courant_global = config->method_courant_global;
 
-  method_list_.push_back(new MethodNull(config->method_null_dt)); 
+  // todo: clean the following up
```
gregbryan, line 471: Maybe make it clearer what is to be cleaned up.

gregbryan (changes requested):
This looks good to me -- the ParameterAccessor seems good to me. Requested changes are just cosmetic. Thanks @mabruzzo!

gregbryan:
Commenting briefly on the questions raised: (1) I think the ability to specify an absolute and relative parameter path seems useful (but if this introduces too much more complexity in the code base, it is less appealing). The parameter-path analogy is immediately clear (at least to me). (2) I *very* much like the idea of eventually introducing checking for parameter use -- as you note, typos in parameter files are an ongoing issue and helping to catch them would be useful.  Again, the issue is how much machinery is required to implement.

jobordner:
Here's a link to my possibly excessive [PR Review](http://client64-225.sdsc.edu/~bordner/review-341.pdf) Nothing really critical--I will accept the PR after you've made whichever of the changes you decide to make. Let me know if you have any questions, and hope you don't mind the unconventional review approach. As the first in a series of PR's, I wanted to err on the side of thoroughness.

mabruzzo:
Thanks James! I'll make a point of addressing these suggestions (they're all really helpful!)

doc/source/devel/param.rst
```diff
@@ -0,0 +1,354 @@
...
+     }
+  }
+
+The organization of parameters in a group hierarchy is analogous to the organization of files in a directory hierarchy.
+Continuing this analogy, we have devised shorthand for naming parameters in the documentation (and throughout the codebase) that is similar to a file path.
+One can think of these names as a "parameter-path".
+
+- documentation refering to the parameters in the above snippet would mention :par:paramfmt:`Method:list`, :par:paramfmt:`Method:mhd_vlct:courant`, or :par:paramfmt:`Method:grackle:courant`.
+
+- in the codebase, one could see strings refering to ``"Method:list"``, ``"Method:mhd_vlct:courant"``, ``"Method:grackle:courant"`` (in practice, you may not see these particular strings).
```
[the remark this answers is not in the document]
mabruzzo: good point

doc/source/devel/param.rst
```diff
@@ -0,0 +1,354 @@
...
+
+
+==============================
+Design Overview (new approach)
+==============================
+
+Our new approach revolves around the usage of the :cpp:class:`!ParameterAccessor` class for accessing/querying parameters stored in an instance of the :cpp:class:`Parameters` class.
+Instances of the :cpp:class:`!ParameterAccessor` class are light-weight and are expected to have a short-lifetime (akin to :cpp:class:`!Field` or :cpp:class:`!Particle`).
+
+**As illustrated above, instances of the** :cpp:class:`!ParameterAccessor` **class are expected to be passed to the constructor (or factory-method) of classes that inherit from Cello class-hierarchy.**
```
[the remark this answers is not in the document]
mabruzzo: Good point! It's not illustrated at all. I added an example

mabruzzo:
> Commenting briefly on the questions raised: (1) I think the ability to specify an absolute and relative parameter path seems useful (but if this introduces too much more complexity in the code base, it is less appealing). The parameter-path analogy is immediately clear (at least to me). (2) I _very_ much like the idea of eventually introducing checking for parameter use -- as you note, typos in parameter files are an ongoing issue and helping to catch them would be useful. Again, the issue is how much machinery is required to implement.

Thanks for responding to these points. It's my plan to definitely to eventually implement parameter-checking (a lot of the necessary infrastructure is already in place from a long time ago). I think I will probably implement the absolute parameter-path stuff at some (assuming it doesn't involve introducing too much machinery). But I probably want to wait on that until after we can remove some functionality from the `Parameters` class (which requires us to migrate a bunch of stuff first).

mabruzzo:
I've finished revising the code. @gregbryan and @jobordner, I'm ready for you to take one more look (I think this about ready to be merged at this point). I haven't changed anything too dramatic (it's mostly just systematic simple changes)

@jobordner in particular, I just wanted to briefly go down your list of suggestions:

1. rename `ParameterAccessor`
	- Per your recommendation, I changed the name to `ParameterGroup`. The new name is more informative and easier to understand.
	- The name is a little misleading/confusing. Because `Param` and `Parameters` instances "own" the associated data, I'd naively assume a `ParameterGroup` instance owns all data in a parameter-group. Instead, it's really a lightweight reference/proxy for a parameter-group.
	- Some alternative names might be `ParamGroupProxy` or `ParamGroupRef` (I don't strongly prefer these alternatives over `ParameterGroup`, since they are a little verbose).
	- I feel like we could spend a long time bikeshedding on this topic. So I'm just going to go ahead and use `ParameterGroup` unless you (or Greg) tell me you prefer one of these alternatives (or a different alternative).
2. prefer a consistent Method creation approach
	- I modified the creation of all Methods in this PR, so that they are all created by passing `ParameterAccessor` to the constructor
	- With that said, I don't think it should be a hard and fast rule that we can't use a factory method.
		- In cases like `EnzoMethodStarMaker`, `EnzoMethodAccretion`, and `EnzoMethodFeedback` where different configurations correspond to different to subclasses it makes a lot of sense to use a factory. (As an aside, it's my intention to eventually refactor those classes - I don't think we should be using 2 levels of inheritance in those cases... I think it produces very confusing code)
		- I can easily imagine a factory-method helping in certain cases where the internal representation of a Method object does not have a one-to-one mapping with the parameters in a parameter file.
3. I agree - this seems like a great idea for the future.
4. rename `get_root_parpath()`
	- Good idea - I did that.
5. remove or rename `wrapped_Parameters_ref()`
	- Good point. I removed it.
6. remove `ParameterAccessor` comments
	- I removed the highlighted comment
7. pass `ParameterAccessor` by value or const reference
	- Good point.  We now pass by value everywhere. It might be better to pass by const-reference, but we would need to declare all of `value` methods of `Parameters` to be const-qualified. But that's a little beyond the scope of this PR.
8. That's a good point. I edited the documentation quite a bit
        - I fleshed out the first half of the documentation a lot more (and gave more examples). I think it should generally be more approachable
        - There's still a lot of room for improvement! In particular, I did not split design docs off from the developer docs. I definitely agree with your point that some of this stuff is better off as design docs, but I don't have it in me to do that right now. 
        - It's my intention to continue improving the docs in the future, but I think it's probably good enough for right now

doc/source/devel/param.rst
```diff
@@ -0,0 +1,451 @@
...
+
+   - ``p.value_float(s, 2.0)`` if the parameter is expected to be a floating-point value and if it defaults to a value of ``2.0`` when the parameter is not specified.
+
+   - ``p.value_string(s, "N/A")`` if the parameter is expected to be a string and if it defaults to a value of ``"N/A"`` when the parameter is not specified.
+
+If the user specified the desired parameter, but with an unexpected type, the program will abort with an error message (another function is provided to query the parameter type)
+
+In each of these snippets, ``s`` is **always** a parameter path, but the precise interpretation depends on how ``p`` is defined.
+When ``p`` references a :cpp:class:`!Parameters` instance, ``s`` must specify an absolute parameter-path.
+In contrast, when ``p`` references a :cpp:class:`!ParameterGroup` instance, ``s`` must specify the path relative to a
```
gregbryan, line 98: This sentence is missing something.

doc/source/devel/param.rst
```diff
@@ -0,0 +1,451 @@
...
+ ::
+
+  Physics {
+    fluid_props {
+      eos    {  gamma = 1.6666666666666667; }
+      floors {  density  = 1.0e-10;         }
+    }
+  }
+
+We will now highlight what the choices of ``s`` that specify is :par:paramfmt:`Physics:fluid_props:eos:gamma` for different choices of ``p``.
```
gregbryan, line 112:
```suggestion
The table below clarifies the value of ``s`` that should be used to specify :par:paramfmt:`Physics:fluid_props:eos:gamma` for different choices of ``p``.
```

doc/source/devel/param.rst
```diff
@@ -0,0 +1,451 @@
...
+  Method {
+     list = [ "heat"];
+
+     heat {
+        alpha = 0.6;
+        courant = 0.3;
+     }
+  }
+
+Here's we present (a heavily editted) example of what the class's constructor might look like:
```
gregbryan, line 157:
```suggestion
Here's we present (an edited) example of what the class's constructor might look like:
```

doc/source/devel/param.rst
```diff
@@ -0,0 +1,451 @@
...
+   We could consider extending the analogy between a parameter-path and a file path.
+   For example, one could imagine interpreting a path that begins with a ``:`` as an absolute parameter-path and all other strings as relative parameter-paths.
+
+   This would probably streamline the documentation to some degree.
+
+   If we were to do that, we would need to modify the code to recognize this convention.
+   We would probably also want to modify the various parameter-accessor methods of the :cpp:class:`!ParameterGroup` to continue to restrict access to parameters within the common root-path that a :cpp:class:`!ParameterGroup` is configured with.
+
+=================
+Historic Approach
```
gregbryan, line 414:
```suggestion
Historical Approach
```

gregbryan (approved):
Modified version looks good to me -- left a few minor comments (but approved because this doesn't need me to look at it again).  Thanks!

PR #346, opened 2023-08-08
Resolve #292 - Drop the ``"Method:flux_correct:single_array"`` parameter

### LINKED ISSUE #292: Dropping the `"Method:flux_correct:single_array"` parameter?
While attempting to refactor the approach for how parameters are used to configure method objects, I came across a quirky parameter: `"Method:flux_correct:single_array"`. 

### Overview

For reference, the value of this method is currently stored within the `method_flux_correct_single_array` attribute of the `Config` object, isn't stored within `MethodFluxCorrect`, and has a default value of `true`.  The quirkiness manifests in the scenario where we configure Enzo-E to invoke the `"flux_correct"` method multiple times per cycle. From a certain perspective, the different  `MethodFluxCorrect` share a common value of `method_flux_correct_single_array`.

While I'm happy to come up with a robust solution that maintains this behavior, but stores the value somewhere other than  the `Config` object, I wanted to ask: **"can we just drop this parameter?"**

### How the the parameter is currently used

As a brief reminder, this parameter is currently only used in 2 primary contexts:
1. This parameter is primarily used within `EnzoMethodMHDVlct` and `EnzoMethodPpm` during calls to `FluxData::allocate`. It controls the memory allocation strategy for the buffers needed to store the flux data necessary to perform a flux correction on a given block (2 buffers are needed per corrected field - 1 holds fluxes computed locally and the other holds data from neighbors). Specifically we either:
    - use a single large memory allocation (and manually break it into the buffers).
    - separately allocate data for each buffer
2. This parameter is also forwarded by `EnzoMethodPpm` to `EnzoBlock::SolveHydroEquations`. If the parameter is `false`, `SolveHydroEquations` knows that it needs to pass dummy arrays to the ppm fortran routines for storing flux-correction data (since the fortran routines currently require the single allocation approach).

### Summary:

If we do decide to drop this parameter, I would modify the codebase to:
- always allocate the flux data buffers as a single array in `EnzoMethodPpm` (since it's a hard requirement of the solver)
- probably do the same in `EnzoMethodMHDVlct` (since it's a well tested codepath). Alternatively, we could add a compile-time parameter to control this.

@jobordner - I'm curious to hear your thoughts about this. Do you feel strongly one way or the other?

mabruzzo:
During the workshop back in May, I had a short discussion with @jobordner about this issue. I believe we decided that this parameter was no longer necessary (thus, I'm removing it).


bwoshea (approved):

bwoshea:
@mabruzzo this looks good and only requires one reviewer to merge.  Could you update the PR so it's up-to-date with `main`?  After the tests pass I will merge.

PR #348, opened 2023-08-14
Migrate Parameter Parsing: migrating configuration of ``EnzoMethodBackgroundAcceleration``

mabruzzo:
Followup from https://github.com/enzo-project/enzo-e/pull/341: moved parameter parsing for ``EnzoMethodBackgroundAcceleration`` from ``EnzoConfig`` into Constructor. 
- this does quite a bit of additional refactoring of ``EnzoMethodBackgroundAcceleration``
    - Namely, I created the ``EnzoPotentialConfigGalaxy`` and ``EnzoPotentialConfigPointMass`` classes, which are just glorified structs. The sole purpose of these structs are to store the configuration parameters for the different background potentials, provide serialization routines, and provide functionality to convert from the input-units to the code units.
    - The fact that we support running cosmological simulations with a background potential makes the implementation slightly more potential (we can't just put the configuration parameters into code-units after reading them in since code-units change every cycle in cosmological simulations)
- This also builds on PR https://github.com/enzo-project/enzo-e/pull/340 (and can't be merged until after that PR).


src/Enzo/gravity/EnzoMethodBackgroundAcceleration.cpp
```diff
@@ -11,281 +11,230 @@
...
-  } else if (enzo_config->method_background_acceleration_flavor == "PointMass"){
-    this->PointMass(ax, ay, az, &particle, rank,
-                    cosmo_a, enzo_config, enzo_units, enzo_block->dt);
-  } else {
-
-    ERROR("EnzoMethodBackgroundAcceleration::compute_()",
-          "Background acceleration flavor not recognized");
+  /// compute the x,y,z components of the acceleration
+  ///
+  /// @note
```
clairekope, line 77: Setting `accel_R` and `accel_z` to 0 for fluids was introduced by [Andrew Emerick back in 2019](https://github.com/enzo-project/enzo-e/commit/147d242a5d95352ed7877ed0843f586d3e2a6c40). According to the commit message, the intention was to create a more stable gas disk. I suspect he was attempting to dodge issues with immediately collapsing disk ICs, which is a problem I have struggled with for isolated galaxies in Enzo as well.

This is, in my opinion, a bad hack that could have terrible unforeseen consequences for anyone using Enzo-E that doesn't expect this to be present. For example, if I were to copy my isolated galaxy CGM sims over to Enzo-E, the gas of the entire CGM would also be affected by this hack and therefore wouldn't infall onto the disk correctly. If I had not reviewed this PR, this would be a challenging issue to debug.

If we want to keep this option for galaxy simulations, it should be parameterized. Otherwise I suggest it be removed and fluids and particles should be impacted by gravity equally.

src/Enzo/gravity/EnzoPotentialConfig.hpp
```diff
@@ -0,0 +1,154 @@
...
+  {
+    return
+      { pack_dfltU.DM_mass * enzo_constants::mass_solar / units->mass(),
+        pack_dfltU.DM_mass_radius * enzo_constants::kpc_cm / units->length(),
+        pack_dfltU.stellar_r * enzo_constants::kpc_cm / units->length(),
+        pack_dfltU.stellar_z * enzo_constants::kpc_cm / units->length(),
+        pack_dfltU.stellar_mass * enzo_constants::mass_solar / units->mass(),
+        pack_dfltU.bulge_mass * enzo_constants::mass_solar / units->mass(),
+        pack_dfltU.bulgeradius * enzo_constants::kpc_cm / units->length(),
+        pack_dfltU.amom,
```
clairekope, line 103: Does `amom` get used as a unit vector?

clairekope (changes requested):
Overall, looks good to me. A very useful refactor that makes it easier to add alternative potentials and, as always, good code style & comments for easy reading :) A few small questions that don't need to be addressed, but I'd like to see the differential treatment of gas and particles by `GalaxyModel` in `EnzoMethodBackgroundAcceleration` addressed before merging. While it wasn't introduced by this refactor, I think the behavior is problematic to preserve.

jwise77 (approved):

src/Enzo/gravity/EnzoPotentialConfig.hpp
mabruzzo: Yes, I have renamed it accordingly

src/Enzo/gravity/EnzoPotentialConfig.hpp
```diff
@@ -0,0 +1,154 @@
...
+  double rcore;
+
+  static EnzoPotentialConfigGalaxy from_parameters(ParameterGroup p) {
+    double DM_mass = p.value_float("DM_mass",-1.0);
+    double DM_mass_radius = p.value_float("DM_mass_radius", 0.0);
+    double stellar_r = p.value_float("stellar_scale_height_r", 1.0E-10);
+    double stellar_z = p.value_float("stellar_scale_height_z", 1.0E-10);
+    double stellar_mass = p.value_float("stellar_mass", 0.0);
+    double bulge_mass = p.value_float("bulge_mass", 0.0);
+    double bulgeradius = p.value_float("bulge_radius", 1.0E-10);
```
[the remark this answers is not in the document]
mabruzzo: fixed

src/Enzo/gravity/EnzoMethodBackgroundAcceleration.cpp
mabruzzo: Fixed!

**Thanks for catching this and bringing this up!** I thought that this was a little funky way back when I first refactored the code. But, I didn't have much experience running simulations with gravity or simulations of galaxies (so I assumed that maybe I was missing something) and then I totally forgot about this.

Since then, I have done a lot of work on self-gravitating isolated galaxies in Cholla. In fact, I spent the entirety of this past February dealing with debugging gravitational instabilities in a self-gravitating gas disk (a little time was spent on vertical instability, but most of the time was related to radial instabilities).

**I wholeheartedly agree** that this hack is detrimental. I have removed the code entirely from the codebase.

mabruzzo:
@clairekope, I finally got around to looking at your review. THANK YOU for taking such a careful look at the code. I completely agree that the "hack" that I had carried over was detrimental and I have removed it from the codebase.

When you have a chance, would you mind signing off on this PR?

clairekope (approved):

PR #349, opened 2023-08-17
Param prototype - first batch of initializers

mabruzzo:
Followup from https://github.com/enzo-project/enzo-e/pull/341: moved parameter parsing the first batch of problem-initializers from  `EnzoConfig` into constructors.

This includes `EnzoInitialInclinedWave` and `EnzoInitialShockTube`.


jwise77 (approved):
This is the first PR I'm reviewing with the ParameterGroup. It's a great addition and makes passing parameters much easier and cleaner. Thanks!

clairekope (approved):

PR #362, opened 2024-02-19
Hilbert order method for load balancing 

brenjohn:
A `MethodOrderHilbert` class was added which uses a Hilbert curve to order blocks in a simulation.

The class is largely based on the MethodOrderMorton class but with some key logic/functions replaced with Hilbert versions. Namely, the logic for iterating over child blocks in a Morton order was replaced with a Hilbert version (facilitated by a `hilbert_children` function) and the `Index.next` method was replaced with `MethodOrderHilbert::hilbert_next`.

The algorithms for determining the Hilbert order are based on those presented in "Mengjuan Li et al 2023 (Efficient entry point encoding and decoding algorithms on 2D Hilbert space filling curve)" and "Lianyin Jia et al 2022 (Efficient 3D Hilbert curve encoding and decoding algorithms)" which are iterative and use look-up tables rather than recursive. Note, there are typos in the tables presented in these papers, the tables used by the `MethodOrderHilbert` class should be the correct ones.

### DIFF
--- a/src/Cello/problem_MethodOrderHilbert.cpp
+++ b/src/Cello/problem_MethodOrderHilbert.cpp
@@ -40,8 +40,8 @@ MethodOrderHilbert::MethodOrderHilbert(int min_level) throw ()
   is_next_         = cello::scalar_descr_index()->new_value(name() + ":next");
   is_weight_       = cello::scalar_descr_long_long()->new_value(name() + ":weight");
   is_weight_child_ = cello::scalar_descr_long_long()->new_value(name() + ":weight_child",n);
-  is_sync_index_  = cello::scalar_descr_sync()->new_value(name() + ":sync_index");
-  is_sync_weight_ = cello::scalar_descr_sync()->new_value(name() + ":sync_weight");
+  is_sync_index_   = cello::scalar_descr_sync()->new_value(name() + ":sync_index");
+  is_sync_weight_  = cello::scalar_descr_sync()->new_value(name() + ":sync_weight");
 }
 
 //======================================================================
@@ -113,8 +113,6 @@ void MethodOrderHilbert::send_weight(Block * block, int weight_child, bool self)
     int na3[3];
     cello::simulation()->hierarchy()->root_blocks(na3,na3+1,na3+2);
 
-    // TODO: use a "Hilbert_next" function here.
-    // Index index_next = block->index().next(rank,na3,block->is_leaf(),min_level_);
     Index index_next = hilbert_next(block->index(), rank, block->is_leaf(), min_level_);
 
     *pindex_(block) = 0;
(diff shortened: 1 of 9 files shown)


src/Cello/problem_MethodOrderHilbert.cpp
```diff
@@ -0,0 +1,556 @@
...
+{
+  // Initialize counters, then barrier to ensure counters initialized
+  // before first entry method can arrive
+  TRACE_ORDER_BLOCK("compute",block);
+  Sync * sync_index = psync_index_(block);
+  Sync * sync_weight = psync_weight_(block);
+
+  *pindex_(block) = 0;
+  *pcount_(block) = 0;
+  *pweight_(block) = 1;
```
jwise77, line 59: Are all blocks equally weighted?  If so, this is fine.  It's just for my curiosity.  If we were to have performance timing for each block or some proxy, would this be where the associated weight would be assigned?

jwise77 (approved):
All looks good! It's more complex than what I did with Enzo's Hilbert curve because you considered the octree structure in its calculation. I had one question for my own curiosity about the block weighting as an inline comment and another question about testing. But I can approve it because those are just minor items.

jobordner:
Looks good to me, and works great!  (out-performs Morton as expected)

In PR #350 I made some structural changes to ordering (e.g. replaced `MethodOrderMorton` with `MethodOrder`), so I'll be merging `MethodOrderHilbert` into that at some point. Just so you know. 

(@jwise77 the weighting is the same as in Morton (the same for all blocks), but the variable "weight" is available so different blocks can be assigned different weights later as needed.)

jobordner (approved):

test/CMakeLists.txt
```diff
@@ -293,3 +293,6 @@ add_custom_target(process_test_results
                   pandoc -s -o TEST_RESULTS.html --metadata title="Enzo-E Test Results" TEST_RESULTS.md
                   WORKING_DIRECTORY ${PROJECT_BINARY_DIR})
 
+# Load Balance
+#setup_test_parallel(LoadBalance-1 LoadBalance/morton-1  input/LoadBalance/test_balance-on-bcg.in)
```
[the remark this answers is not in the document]
brenjohn: The test for the Morton method was actually already removed before I added the Hilbert method. My guess is the original test was removed because it was a lengthy cosmological simulation. I added it back in temporarily while I was putting together the Hilbert test and then commented it out. I didn't want to reintroduce a test which was removed and I probably should have removed it entirely instead of commenting it out.

src/Cello/problem_MethodOrderHilbert.cpp
brenjohn: Yeah, the weighting is the same as in the Morton method as @jobordner said. So all the blocks initially have equal weight and by the end of the method all leaf blocks should have weight 1 and all non-leaf blocks should have a weight = 1 + (the number of its descendant blocks)

PR #370, opened 2024-02-28
Address lex.yy.c compiler warnings

mabruzzo:
### Pull request summary

I made 2 tweaks to the lexer code in order to address compiler warnings

### Detailed Description

The specific changes include:
- I addressed the warning that the yyunput function is defined, but unused
by adding the nounput option to parse.l
- I addressed compiler warnings about strncpy the case when we try to
  copy a line of text into our line buffer, but the line is larger than our
  line buffer.
  - I vaguely recall that there may be some safeguards in other areas of
    the code that should stop the program before we execute the code that
    the compiler was warning about.
  - But just in case we do encounter this scenario, I added logic to
    print out a warning to explain to users what is happening. Even if
    this code is never executed, it should make things a little more clear
    for future readers of `parse.l`.

Finally, I used flex to regenerate `lex.yy.c` from `parse.l`


jobordner (approved):
Looks good, free to merge.

PR #372, opened 2024-03-18
Drop Boost as Dependency

mabruzzo:
### Pull request summary

This PR drops `boost::filesystem` as a dependency (now that we are using C++ 17, we can use `<filesystem>` instead, which is shipped as part of the standard library).

### Detailed Description

This change marginally reduces Enzo-E's installation requirements (I think it's generally a good thing).

In the build system, I:
- added some basic error-checking to the build-system to inform the user if there is an issue with `std::filesystem`
- also tried to add more complete support for older versions of the standard library implementations (from before ~July 2019).  In these cases you needed to pass `-lstdc++fs` or `-lc++fs` to the linker, depending on whether you're using the implementation from GNU or LLVM (this is analogous to how you need to pass `-lm` if you want to use function from `<math.h>` in a C program).

Honestly, these build-system changes are a little over-engineered (and probably unnecessary). 99% of people shouldn't encounter any issues. An argument could be made for removing these build-system changes (particularly the latter change).

I don't think this is a major change (it probably doesn't need 2 reviewers). 

### PR Checklist

Note that some of these check boxes may not apply to all pull requests.


brenjohn (approved):
This looks good to me! Just needs to be updated with latest changes in main.

PR #374, opened 2024-03-18
Migrate Parameter Parsing: migrating configuration of `EnzoMethodMHDVlct`

mabruzzo:
### Pull request summary

Followup from #341: moved parameter parsing for `EnzoMethodMHDVlct` from `EnzoConfig` into constructor (and helper functions).

### Detailed Description

This is part of the large migration of Parameter Parsing (that has been split into multiple parts to ease the entire process). The benefits of this migration are highlighted in the user-documentation on the website.

### PR Checklist

Note that some of these check boxes may not apply to all pull requests.


src/Enzo/hydro-mhd/EnzoMethodMHDVlct.cpp
```diff
@@ -35,12 +35,86 @@ static str_vec_t concat_str_vec_(const str_vec_t& vec1, const str_vec_t& vec2){
...
+      // algorithm to use anything other than the "nn" choice. (The only reason
+      // it was ever an option was due to a misunderstanding early on)
+      ERROR1("parse_mhdchoice_pack_pair_",
+             "The deprecated parameter, \"%s:half_dt_reconstruct_method\", "
+             "can't have any value other than \"nn\"",
+             p.get_group_path().c_str());
+    }
+
+    if (is_defined("full_dt_reconstruct_method")) {
+      recon_names[1] = p.value_string("full_dt_reconstruct_method", "plm");
```
clairekope, line 86: Is assuming `recon_names` has two entries going to be an issue? e.g. is it possible for the `time_scheme` to be Euler and reach this piece of code? Doesn't seem so, as `time_scheme` must be undefined to reach this block and it's default value is `vl`, but wanted to double check :)

src/Enzo/enzo-core/EnzoProblem.cpp
```diff
@@ -686,13 +686,8 @@ Method * EnzoProblem::create_method_
...
 
-    method = new EnzoMethodMHDVlct
-      (enzo_config->method_vlct_riemann_solver,
-       enzo_config->method_vlct_time_scheme,
-       enzo_config->method_vlct_reconstruct_method,
-       enzo_config->method_vlct_theta_limiter,
-       enzo_config->method_vlct_mhd_choice,
-       store_fluxes_for_corrections);
+    method = new EnzoMethodMHDVlct(p_group, store_fluxes_for_corrections);
+    skip_auto_courant = true;
```
clairekope, line 690: I don't see any other references to this variable in this PR. What is its purpose?

src/Enzo/initial/EnzoInitialShuCollapse.cpp
clairekope, line 1: Is the functionality of these ASSERTs captured elsewhere? Or is there a reason for removing them? This comment also applies to the two other initial condition files that had ASSERT statements removed

clairekope (approved):
I have a few small questions about the changes made in this PR but otherwise I approve

src/Enzo/hydro-mhd/EnzoMethodMHDVlct.cpp
mabruzzo: Thanks for pointing this out! 

I definitely could have made a mistake here, but I think this is ok:
- On [line 46](https://github.com/enzo-project/enzo-e/blob/ba857eef6d8f78958915fd074dddb3d46d9f994f/src/Enzo/hydro-mhd/EnzoMethodMHDVlct.cpp#L46), logic is in place that assigns the `time_scheme` variable a value of `"vl"` if the `Method:mhd_vlct:time_scheme` parameter was undefined.
- The logic here is indeed a little convoluted (basically, I wanted to force people to stop using the deprecated parameters if they started using the newer parameters)

I think I'll submit a new PR to remove this deprecated parameter completely. I was tempted to do that sooner (since I'm the main/only user of this solver right now), but I didn't want to keep the removal out of this PR.

src/Enzo/initial/EnzoInitialShuCollapse.cpp
mabruzzo: Good catch! I don't know why I removed the mhd-check. It should definitely be there. I'll modify this PR so that we perform this check.

With that said, I don't think the Riemann Solver check is necessary. In fact, I am very hesitant to expose that piece of information to arbitrary methods and initializers unless there is a very compelling reason (since we know we'll be using pure hydro in this case, that should really be unnecessary here).

clairekope: Whatever you think makes the most sense! I just want to be sure we're consciously choosing to "downgrade" hard asserts to soft code comments.

src/Enzo/hydro-mhd/EnzoMethodMHDVlct.cpp
clairekope: I agree the current program flow doesn't look like it should introduce any errors (at least from what I see within this PR) so it's fine if you want to refactor in another PR. It's probably best to keep this PR a more straightforward port.

src/Enzo/enzo-core/EnzoProblem.cpp
mabruzzo: Enzo-E has traditionally read and set the courant factor of every method automatically. However, we are moving away from that. Now, Methods read in their own courant factor (since different methods have different requirements - e.g. it should never exceed 0.5 for vl+ct). This variable is being used to indicate that courant factor should not automatically be set

PR #375, opened 2024-03-19
we dropped a handful of `ifdef` statements involving `CONFIG_USE_GRACKLE`

mabruzzo:
### Pull request summary
This drops a handful of `ifdef` statements involving `CONFIG_USE_GRACKLE`

### Detailed Description

The benefits are pretty self-explanatory -- namely, it simplifies the codebase. In this particular case, it also simplifies some of the build files. It's also a step towards my goal of removing all of the `CONFIG_USE_GRACKLE` from all headers and just have it in a handful of source files in Enzo's `chemistry` subcomponent (so that rebuilding Enzo-E with/without grackle support can be done as incremental recompilation)


WillHicks96 (approved):
Looks good. There are merge conflicts in `EnzoMethodMHDVlct.cpp`

gregbryan (approved):

PR #378, opened 2024-03-26
Migrate Parameter Parsing: moved configuration of `EnzoInitiailFeedbackTest` into constructor

mabruzzo:
### Pull request summary

Followup from #349: moved parameter parsing for `EnzoInitiailFeedbackTest` from `EnzoConfig` into constructor.

**Aside:** This sort of migration is the sort of thing that needs to be done for a lot more examples. This is a great first-time-contribution for new developers. I'm going to create an Issue summarizing the process and that links to this PR as an example (in reality, I will probably link to the particular commit that makes this change).

### DIFF
--- a/src/Enzo/tests/EnzoInitialFeedbackTest.cpp
+++ b/src/Enzo/tests/EnzoInitialFeedbackTest.cpp
@@ -18,14 +18,32 @@
 int nlines(std::string fname);
 
 
-EnzoInitialFeedbackTest::EnzoInitialFeedbackTest
-(const EnzoConfig * config) throw ()
-  : Initial(config->initial_cycle, config->initial_time)
+EnzoInitialFeedbackTest::EnzoInitialFeedbackTest(int cycle, double time,
+                                                 ParameterGroup p) throw ()
+  : Initial(cycle, time),
+    density_CGS_(p.value_float("density", 1.0E-24)),
+    temperature_(p.value_float("temperature", 1.0E4)),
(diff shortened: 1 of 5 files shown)


jwise77 (approved):

src/Enzo/tests/EnzoInitialFeedbackTest.cpp
```diff
@@ -143,32 +164,40 @@ void EnzoInitialFeedbackTest::enforce_block
   int ngz = nz + 2*gz;
 
   const enzo_float gamma = enzo::fluid_props()->gamma();
+  // shouldn't we be using a self-consistent mmw?
   const enzo_float mol_weight = enzo::fluid_props()->mol_weight();
```
clairekope, line 168: I assume this is reading the molecular weight from somewhere else, like another parameter? Maybe make this an issue so we don't loose track of it?

clairekope (approved):
One question following up on a code comment, but this has nothing to do with the functionality of the PR

src/Enzo/tests/EnzoInitialFeedbackTest.cpp
mabruzzo: Yeah, users have the option to specify a single,fixed mmw weight by using the parameter `Physics:fluid_props:mol_weight`. I added the comment because we go on and subsequently use various species fields. 

It doesn't appear that the mmw computed from the species fields will be consistent with the mmw specified in the parameters.

(I'll add this note to the species issue)

PR #381, opened 2024-04-10
Add gravity method super-cycling

jobordner:
Add super-cycled gravity, and support for super-cycling other methods.

This PR adds support for super-cycling methods, while still always satisfying method-specific timestep constraints. The purpose is to improve runtime performance, which can be especially helpful when one specific method (e.g. gravity) tends to run slower than other methods. This PR implements super-cycling for the gravity method in particular, and provides helper methods to implement super-cycling for other methods.

 The time-step it uses for a super-cycled method is determined using 1. a "`max_supercycle`" parameter defined in the `Method : <method>` group, and 2. the ratio of time-steps for the super-cycled method relative to the global minimum. This allows the user to control the maximum "amount" of super-cycling, while never having to worry about violating time-step constraints.

Super-cycling works by allowing the Method to have a "solve" cycle, followed by some number k of "non-solve" cycles 0 <= k < `max_supercycle`. During "solve" steps, the method saves some extra fields, which are used in "non-solve" cycles to approximate the fields normally computed by the method. For the gravity method in particular, these saved fields may be either the gravitational potential, or the acceleration fields, as determined by the "`Method:gravity:type_supercycle`" parameter.

- New features are documented with narrative docs: see `doc/source/design/design-timestep.rst`
- Adds tests for new features: see `input/test_cosmo-dd8.in`

- [ X] This is a major change or addition.  Will require two reviewers.


doc/source/design/design-timestep.rst
```diff
@@ -0,0 +1,202 @@
...
+ENZO code: ENZO uses different time-steps on different mesh refinement
+levels, which can significantly improve performance on mesh
+hierarchies with many refinement levels. An example of 2) would be
+advancing the gravitational potential, which can be costly due to its
+global communication requirements, at a larger time-step than other
+methods when possible.
+
+Our plan is to optimize time-stepping in terms of both methods (via
+"super-cycling") and individual blocks (via "block-adaptive
+time-stepping").  We implement super-cycling first since it requires
```
mabruzzo, line 45: Does block-adaptive time-stepping mean that blocks on the same refinement level can have different timesteps?

doc/source/design/design-timestep.rst
```diff
@@ -0,0 +1,202 @@
...
+methods. The default value of :code:`1` effectively turns off
+super-cycling, whereas a value of :math:`k` allows other methods to be
+called *up to* :math:`k` times for each call to the gravity
+solve. Note that the maximum local gravity time-step restriction on
+each block will still always be satisfied, even if the
+:code:`max_supercycle` parameter is set to an arbitrarily large value.
+
+For cycles where the gravitational potential solve is not performed,
+extrapolations of saved fields are used to approximate the current
+fields.  Which fields to be extrapolated is determined by the
```
mabruzzo, line 74:
```suggestion
fields.  The fields to be extrapolated are determined by the
```

config/linux_gcc.cmake
```diff
@@ -17,6 +22,12 @@ if(NOT __processedUserDefaults)
   # Note (12/2021): passing -march=native to gfortran seems to slow down the
   # PPM solver
 
+  set(__ARCH_C_OPT_FLAGS "-O3 -g -funroll-loops")
+  set(CMAKE_C_FLAGS_RELEASE "${__ARCH_C_OPT_FLAGS}")
+  set(CMAKE_C_FLAGS_RELWITHDEBINFO "-g ${__ARCH_C_OPT_FLAGS}")
+  set(CMAKE_CXX_FLAGS_RELEASE "${__ARCH_C_OPT_FLAGS}")
+  set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "-g ${__ARCH_C_OPT_FLAGS}")
+
```
mabruzzo, line 30:
```suggestion
```

I think these are now all unnecessary due to some more recent changes.

If you look further down on the page, we define `ENZOE_C_FLIST_INIT` and `ENZOE_CXX_FLIST_INIT`. 
```c++
  set(ENZOE_C_FLIST_INIT "-Wall;-funroll-loops")
  set(ENZOE_CXX_FLIST_INIT "${ENZOE_C_FLIST}")
```
Currently, these add the `-Wall` and `-funroll-loops` flags. If you also want the `-O3` flag instead of `-O2`, you should modify the list so that it reads `-Wall;-funroll-loops;-O3`.

As an aside, we shouldn't be adding `-g` as a symbol here. The canonical way to get debugger symbols is to use `-DCMAKE_BUILD_TYPE=RelWithDebInfo` when configuring the build-system.

doc/source/design/design-timestep.rst
```diff
@@ -0,0 +1,202 @@
...
+
+The gravitational potential is used to compute gravitational
+accelerations, which are applied to all gravitating particles and
+fields. We use a leapfrog integration scheme because it is 2nd order
+accurate (small short-scale errors) and symplectic (qualitatively
+correct long-scale behavior). This involves computing time-centered
+accelerations. Since we start with non-time centered values, we must
+either extrapolate the input (density) to the gravity solve, or
+extrapolate the output (potential) from the gravity solve. In both
+ENZO and Enzo-E, the extrapolation is done before the gravity solve.
```
mabruzzo, line 110: Extrapolating the density involves the `Method : pm_deposit : alpha` parameter, right?

If so then our approach isn't robust in the general case. When I implemented self-gravity source terms for VL+CT, there was a whole saga where I was getting bizare looking results for a standard test problem (i.e. the "stable linear Jeans wave" that is inclined with respect to the grid and for which we can perform an analytic convergence study). Ultimately, the solution was to stop extrapolating. I suspect that we would find similar issues with PPM if we ran a similar test

In that scenario, there were no particles in that case. It's plausible that if particles dominate the short-range gravitational potential that extrapolation might be more robust.

Anyway, we probably don't want to go into detail here, but maybe we add something like the following at the end of this subsection?

```rst
.. caution:

   Some care is needed when applying extrapolation methods. For certain problems, they may not produce meaningful results.
```

doc/source/design/design-timestep.rst
```diff
@@ -0,0 +1,202 @@
...
+
+.. |tijk| replace:: t\ :sub:`i` :sup:`j,k`
+
+.. toctree::
+
+.. _Adaptive Time-Step Design:
+
+
+*************************
+Adaptive Time-Step Design
```
mabruzzo, line 11: This isn't essential, but we may want to acknowledge that subcycling also exists (e.g. in Grackle and radiative transfer). Maybe just a single sentence noting that this is used as a technique on a method-by-method basis.

Issue #383, opened 2024-04-18
Refactoring Parameter Retrieval


mabruzzo:
Beginning with PR #341, Enzo-E is in the process of refactoring the way that parameters from the parameters file are retrieved and used to configure various classes of objects throughout the Cello and Enzo layers.

The new approach for configuring instances of subclasses of `Method` and `Initial`  is described [here]([https://enzo-e.readthedocs.io/en/latest/devel/param.html](https://enzo-e.readthedocs.io/en/latest/devel/param.html)) in more detail.

We need help with this process. This would be an excellent way for first time contributors to contribute to Enzo-E.

## What Needs to Be Done:

Time permitting, I would love to add more details about the specifics of completing this process. But for now the following 2 links show examples of what needs to be done.

2 Existing Pull Requests exist that demonstrate how this is done. 
- PR #379 illustrates a fairly straight-forward example
- PR #378 shows a more involved example. It demonstrates that we don't want our refactored classes to access the contents of `EnzoConfig`, which can be accomplished with `enzo::config()`  (we also don't want them to access `Config`, which can be retrieved with `cello::config()`)

Guidelines: please have each pull request perform this refactoring on just a single class (that way this is easy to review). A number of classes have already been refactored. Please pick from the list down below (and maybe leave a comment down below so we can make sure nobody duplicates your efforts).

Quick tip: if you encounter an initializer that expects `gamma` (adiabatic index), you should access it with `double gamma = enzo::fluid_props()->gamma();`

## Cases that need to be addressed:

At the moment, this is a somewhat incomplete list (but I plan to update it -- I definitely need to add `Method` classes). I've tried to sort in terms of which will be easier/more straightforward  and which will be more difficult (but, I only took a quick glance so my sorting may be wrong).

More straight-forward cases:
- [ ] `EnzoInitialAccretionTest`
- [ ] `EnzoInitialBBTest`
- [ ] `EnzoInitialCollapse`
- [ ] `EnzoInitialImplosion2` (especially easy: just make `ParameterGroup*` an argument of the constructor; it doesn't need to access any parameter)
- [ ] `EnzoInitialPpmlTest` (especially easy: just replace the `EnzoConfig*` constructor argument with `ParameterGroup`. It doesn't actually need to access any parameters)
- [ ] `EnzoInitialSoup`
- [ ] `EnzoInitialShuCollapse`

Relatively more difficult cases:
- [ ] `EnzoInitialBCenter` -- reserved by @mabruzzo 
[...]

PR #388, opened 2024-05-14
EnzoInitialShuCollapse parameter refactorization

jakereinheimer:
### Pull request summary

Refactoring EnzoInitialShuCollapse parameters. Addresses issue https://github.com/enzo-project/enzo-e/issues/383. Short because I am going through it live with Matthew

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,42 +56,7 @@ if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
     "Debug" "Release" "MinSizeRel" "RelWithDebInfo")
 endif()
 
-find_package(Charm REQUIRED)
-# Link executables with the charmc wrapper
-STRING(REGEX REPLACE "<CMAKE_CXX_COMPILER>" "${CHARM_LINKER}"
-       CMAKE_CXX_LINK_EXECUTABLE "${CMAKE_CXX_LINK_EXECUTABLE}")
-
-
-# Need to process Grackle here as both Cello and Enzo-E depend on it
-option(USE_GRACKLE "Use Grackle Chemistry" ON)
-# don't bother advertising the following option (but users can
-# overwrite it if they really want to - e.g. to reduce binary size)
-option(GRACKLE_USE_STATIC_LIBS  "sets Grackle's lib-type if USE_GRACKLE=ON" ON)
-if (USE_GRACKLE)
-  find_package(Grackle)
-  if (Grackle_FOUND)
-    # This really only needs to be defined for a relatively small subset of
-    # files in the Enzo-layer
-    add_compile_definitions(CONFIG_USE_GRACKLE)
-
-  else()
-    message(FATAL_ERROR
-      "Configured to use Grackle but Grackle library not found.\n"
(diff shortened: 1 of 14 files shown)


jwise77:
There are some compilation errors
```
/home/circleci/enzo-e/src/Enzo/initial/EnzoInitialShuCollapse.cpp:21:4: error: expected ‘{’ before ‘center_’
   21 |    center_{}
      |    ^~~~~~~
/home/circleci/enzo-e/src/Enzo/initial/EnzoInitialShuCollapse.cpp: At global scope:
/home/circleci/enzo-e/src/Enzo/initial/EnzoInitialShuCollapse.cpp:21:4: error: ‘center_’ does not name a type
/home/circleci/enzo-e/src/Enzo/initial/EnzoInitialShuCollapse.cpp:22:4: error: ‘drift_velocity_’ does not name a type
   22 |    drift_velocity_{}
      |    ^~~~~~~~~~~~~~~
/home/circleci/enzo-e/src/Enzo/initial/EnzoInitialShuCollapse.cpp:23:1: error: expected unqualified-id before ‘{’ token
   23 | {
      | ^
```

src/Enzo/initial/EnzoInitialShuCollapse.cpp
```diff
@@ -10,31 +10,24 @@
...
-   instability_parameter_(instability_parameter),
-   external_density_(external_density),
-   central_sink_exists_(central_sink_exists),
-   central_sink_mass_(central_sink_mass)
+   truncation_radius_(p.value_float("truncation_radius",0.0)),
+   nominal_sound_speed_(p.value_float("nominal_sound_speed",0.0)),
+   instability_parameter_(p.value_float("instability_parameter",0.0)),
+   external_density_(p.value_float("external_density",0.0)),
+   central_sink_exists_(p.value_logical("central_sink_exists",false)),
+   central_sink_mass_(p.value_float("central_sink_mass",0.0))
```
mabruzzo, line 20:
```suggestion
   central_sink_mass_(p.value_float("central_sink_mass",0.0)),
```

src/Enzo/initial/EnzoInitialShuCollapse.cpp
```diff
@@ -10,31 +10,24 @@
...
+   center_{}
```
mabruzzo, line 21:
```suggestion
   center_{},
```

mabruzzo (changes requested):
This is fantastic! I really appreciate you doing this!

I highlighted 2 missing commas that are needed in order to get this to compile properly.

There are also 2 other changes that are necessary:
1. The order in which the members appear in the "member-initializer-list" in the constructors should match the order in which they are declared within the class itself (jump to the end of `EnzoInitialShuCollapse` to see this ordering).
    - technically, this is not strictly necessary, but it's good practice (if you don't do this, you can get unexpected behavior while using one member to initialize another member) 
    - plus `g++` raises some annoying warnings if you don't do this
2. More importantly, you are changing the default value of the parameters
    - this is my fault for not explaining this.
    - When you call something like `p.value_float("truncation_radius", ...)` the trailing value represents the default value.
    - Currently, you've changed the default values compared to what was previously used within `EnzoConfig::read_initial_shu_collapse_`

PR #391, opened 2024-05-15
Replacing sprintf

mabruzzo:
### Pull request summary

As mentioned in #390, macOS has decided that `sprintf` is deprecated, which produces a lot of annoying compiler warnings. This replaces all occurrences of `sprintf` outside of the parsing code (which is addressed by #390)

### Detailed Description

This is pretty self-explanatory. I would have loved to have consolidated the implementations of `cello::expand_name` and `Output::expand_name_``, but I was a little confused about the differences (and I was trying to get through this relatively quickly).

Aside: In case you're unaware, if you declare a c-style array
```c
char my_arr[14];
```
then calling `sizeof(my_arr)` gives the total size of the array in bytes (in contrast, calling `sizeof` on a pointer just gives the size of the pointer). I take advantage of this a lot throughout this PR.

This PR has no dependencies. (It is totally independent from #390)

### DIFF
--- a/src/Cello/_data.hpp
+++ b/src/Cello/_data.hpp
@@ -160,9 +160,9 @@ class FieldFace;
     CkPrintf ("DEBUG_FIELD %s %20.18g %20.18g %20.18g\n",  NAME,sum_abs,sum_mean,sum_var); \
     fflush(stdout);							\
     char filename[80];						\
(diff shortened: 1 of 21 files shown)


bwoshea (approved):
Looks good to me!

PR #394, opened 2024-05-16
Suppress spurious linker warnings on macOS

mabruzzo:
### Pull request summary

When compiling on macOS (with the Xcode 15 toolchain), lots of spurious linker-warnings crop up. This addresses those warnings.

### Detailed Description

In more detail, Apple introduced a new linker with their Xcode 15 toolchain (I think they shifted from lld to sold). This triggers lots of warnings when linking multiple static libraries. This is a well-known issue and CMake version 3.29 took steps to fix this issue. 

I spent a ton of time trying to use this proper fix (since if you're on macOS, it's extremely easy to download the newest cmake version). 
- Unfortunately, things are tricky since we are forcing cmake to use `charmc` for linking. 
- `charmc` is a bash script created by charm++ that performs an analogous function to something like `mpicc`,`mpicxx`, etc. In other words, it is a compiler wrapper. 
- Aside: it's frankly a little surprising to me that the way that we currently force cmake to use `charmc` gives the desired result (I elaborate in some documentation I added - I also touch on how we probably "should" handle this case if we ever encounter issues).
- In any event, `charmc` typically wraps a compiler-frontend (like gcc or clang) and uses the wrapped compiler-frontend to invoke the linker. But, `charmc` and the wrapped compiler-frontend employ different sets of flags to escape arguments intended only for the linker. The whole thing is a mess.

It's probably possible to make things work. But, since it wasn't easy to do (and since I'm a little skeptical that the way that we force cmake to invoke charmc isn't truly forward-compatible), I decided to adopt a more crude solution. I tried to document as much of this as possible with comments. 

I also suppressed some warnings about extra unused arguments. I think these are all `charmc` issues (but it a subset of them might be cmake-related -- maybe because it doesn't really understand what linker is being used?). It's plausible future charm-releases might fix these issues...

Finally, the following warning
```sh
ld: warning: -multiply_defined is obsolete
```
still gets printed, but there's definitely a ``charmc`` issue.

----

It would be nice to try compiling against the dev-version of charm++ to see if some of the remaining warnings go away (if not, I could potentially open an issue upstream). Unfortunately, there are currently some issues with doing that (from a quick glance, I think our default choices for load-balancing strategies can't be specified in same way in the newer charm++ version)


cmake/FindCharm.cmake
```diff
@@ -62,15 +66,54 @@
...
+#  While there does not appear to be exact documentation describing exactly how
+#  CMAKE_CXX_LINK_EXECUTABLE, based on most of CMake's generic semantics, there
+#  is reason to be skeptical of this kind of approach.
+#  - In general, cmake expects variables like these to be set in a "toolchain
+#    file" (essentially as early in the cmake startup as possible).
+#  - Unfortunately, this approach is a undesirable since users need to craft
+#    their toolchain files ahead of time (and they can't rely upon the
+#    find_package command)
+#
+#  The convential approach (that we should probably adopt if the current
```
bwoshea, line 110: The parenthetical in this sentence doesn't actually end - you might want to add it?

cmake/FindCharm.cmake
```diff
@@ -62,15 +66,54 @@
...
+#  approach) for handling the case where you want to link a program that makes
+#  use of framework that provides a special compiler-wrapper frontend (like
```
bwoshea, line 112: of A framework

cmake/FindCharm.cmake
```diff
@@ -97,8 +140,103 @@ function(GET_CHARM_QUEUE_TYPE CHARM_RNDQ conv_conf_hdr)
...
+  # ARGS
+  # ----
+  # charm_compiler - the path to charmc
+  # out            - specifies the variable-name that will be modified by this
+  #                  function
+  #
+  # WHY THIS EXISTS
+  # ---------------
+  # This function is not strictly necessary. In all cases where the charm
+  # compiler was build with a consistent set of compilers to the ones used in
```
bwoshea, line 160: build -> built

bwoshea (approved):
All I found was minor typos.  Otherwise it looks fine to me!

Issue #400, opened 2024-05-21
Use ``fmt`` C++ library? OPINIONS WANTED!


mabruzzo:
This is an idea I had a few years back. I've gone back and forth on it a lot (I'm torn about whether the benefits outweigh the negatives... it seems a little close). I figured that I would throw it out there to see what people think

## The problem

Using data to format strings in C++ is unsatisfactory. Historically, there are 2 approaches:
1. use c-style `printf`-functions. (what we currently do)
2. use C++-style streams like `std::cout` or `std::cerr`

Of these 2 options I'm generally a fan of the first since it is far less verbose and the data-formatting syntax is fairly "standard". However, the major drawback is that it can't properly handle any non-C primitives (e.g. it handles `const char*` strings, but not `std::string` or a user-defined custom string implementation).

## The potential solution

The potential solution to is to make use of ``fmt`` C++ library ([docs](https://fmt.dev/latest/index.html) and [GitHub repository](https://github.com/fmtlib/fmt)). This is an extremely popular library that takes a ``printf``-style approach to string formatting. In fact, it is so popular, that it has become part of the standard library (most features were added in C++20, a few more introduced in C++23 and C++26). 

The basic premise is that it performs formatting based on python's `str.format(*arg, **kwargs)` formatting string. As a basic example, if one wrote

```c++
std::string s = fmt::format("This is a float {:.3f} and this is an int {}", 42.39542342, 42);
```

Then ``s`` would hold ``"This is a float 42.395 and this is an int 42"``.

Importantly, it provides:
- support for formatting types in the standard library. This includes containers like `std::vector`. It also includes ``std::string`` (and supporting `std::string` is extremely important)[^1]
- a mechanism to allow formatting of custom types (which could help with debugging).
- a ``fmt::format_to`` function that can be used to avoid heap allocations.

I think a **major** advantage is that in 3 to 6 years (whenever we update to future C++ releases), we could eventually remove ``fmt`` as a dependency and use the standard library instead (just like we did with ``boost::filesystem``).

## Drawbacks

[...]

PR #401, opened 2024-05-21
Introduce `CELLO_ERROR`, `CELLO_REQUIRE`, `CELLO_WARN`

mabruzzo:
### Pull request summary

This introduces 3 function-like macros for error-reporting and logging purposes. The intention is to replace the existing ``ERROR``, ``ASSERT`` and ``WARNING`` families of macros.

#### About ``CELLO_ERROR``

For the sake of motivating this change, let's focus on the `CELLO_ERROR` macro. This macro should be treated as though it is a function with the declaration
```c++
[[noreturn]] void CELLO_ERROR(const char* format, ...);
```

The idea is that this single macro replaces the entire exiting ``ERROR`` macro family. For context, those macros are used as though they have the following function signatures: 

```c++
// in this snippet,  T1, T2, ... are just used as undeclared place-holder types.
[[noreturn]] void ERROR(const char* func_name, const char* format);
[[noreturn]] void ERROR1(const char* func_name, const char* format, T1 arg1);
[[noreturn]] void ERROR2(const char* func_name, const char* format, T1 arg1, T2 arg2);
// skipping ERROR3, ERROR4, ERROR5, ERROR6, and ERROR7
[[noreturn]] void ERROR8(const char* func_name, const char* format, T1 arg1, T2 arg2, T3 arg3, T4 arg4, T5 arg5, T6 arg6, T7 arg7, T8 arg8);
```

Essentially, ``CELLO_ERROR`` makes 2 key innovations:
1. we no longer need to specify ``func_name``, the name of the function that we are invoking the macro from. This is detected automatically by ``CELLO_ERROR``
    - we can detect the function name by using the `__func__` variable that is implicitly define within every function/method starting with c++ 11[^1]
    - this is important. People sometimes currently forget to manually change `func_name` when they copy an error between functions or they rename a function. This change will make things completely unnecessary.
2. We no longer have separate macros for different numbers of formatting arguments.

#### Other macros

Just as ``CELLO_ERROR`` replaces the ``ERROR`` macro family,

- ``CELLO_REQUIRE`` replaces the ``ASSERT`` macro family
- ``CELLO_WARN`` replaces the ``WARNING`` macro family

These other 2 macros should be treated as though they are functions with the following signatures

```c++
void CELLO_REQUIRE(bool satisfies_requirement, const char* format, ...);
void CELLO_WARN(const char* format, ...);
```

One important note: in the ``ASSERT`` macro-family, the assertion condition is always the last argument. In contrast, the assertion-condition is always the first argument of ``CELLO_REQUIRE``.
- This difference is just a consequence of how variadic macros work.
- It's also part of the reason I named the replacement ``CELLO_REQUIRE`` rather than ``CELLO_ASSERT``.[^2]

#### Caveats

- The main caveat is that these new macros will coexist alongside the old macros.
    - From some quick shell commands, it appears that there are ~1000 occurrences of the old macros. I don't really have any interest in systematically going through and replacing the old macros (I could probably be convinced to remove the versions that took more than say 5 arguments, I think that would be tractable)
    - I personally think that the utility of these new commands are significant enough to warrant them existing side-by-side.
    - But if any reviewer (or anybody in general) disagrees, I'm totally fine with that. We can just close the PR.

- In the event that this is accepted, I think we should hold off on merging this until after we make a decision on Issue #400 (to be clear, **I think the review of this PR should proceed without worrying about that other issue** - we may just want to coordinate things if we decide to proceed with both proposals).

This is a major change or addition (that needs 2 reviewers): unknown

### PR Checklist

[^1]: where available, we prefer compiler-specific alternatives like ``__PRETTY_FUNCTION__`` on gcc and clang. This gives a more descriptive name (including the scope, any template specializations, and the precise function signature).
[^2]: the other reason is to make ``CELLO_REQUIRE`` a little more distinct from the ``assert`` macro introduced by ``<assert.h>``. The key point is that the behavior of ``CELLO_REQUIRE`` is completely independent of other macros. In contrast, ``assert`` becomes a no-op when the ``NDEBUG`` macro is defined.

### DIFF
--- a/doc/source/devel/coding-guidelines.rst
+++ b/doc/source/devel/coding-guidelines.rst
@@ -455,23 +455,33 @@ We recommend some of the following coding practices:
        This generally leads to more explicit code.
        Furthermore, it lets the compiler identify cases where the order of an integer argument and a scoped enum argument are accidentally permuted.
 
-* **ERROR REPORTING:** We don't use C++ exceptions in the codebase. In general, when an error arises, we generally abort the program with an informative error message. You can signal that an error occured by using the ``ERROR`` family of macros or you can use the ``ASSERT`` family of macros to have the program conditionally abort if some condition is not satisfied.
+* **ERROR REPORTING:** We don't use C++ exceptions in the codebase. In general, when an error arises, we generally abort the program with an informative error message. You can signal that an error occured by using the :c:macro:`!CELLO_ERROR` function-like macro. Alternatively you can use the :c:macro:`!CELLO_ABORT` function-like macro to conditionally abort if some condition is not satisfied.
 
-  * When you are implementing new functionality, you are encouraged to liberally use the :c:macro:`!ERROR` and :c:macro:`!ASSERT` families of macros to ensure that Enzo-E loudly fails and aborts when the functionality is used in unexpected ways.
+  * When you are implementing new functionality, you are encouraged to liberally use the :c:macro:`!CELLO_ERROR` and :c:macro:`!CELLO_REQUIRE` macros to ensure that Enzo-E loudly fails and aborts when the functionality is used in unexpected ways.
     When running a really expensive simulation, a user should generally prefer that a simulation loudly fails.
     The extreme alternative case is for the simulation to run to completion while silently having problems, which likely invalidates the results (and these problems may not be detected until MUCH later).
     Furthermore, it's easy enough for a user to comment out an error message that they wish to ignore.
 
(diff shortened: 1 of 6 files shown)


jobordner (approved):
Overall very nice!
Would it be difficult to keep CELLO_ASSERT in addition to CELLO_REQUIRE? My original intention was to use CELLO_ASSERT for checking code during development, and allow it to be disabled in production code. CELLO_REQUIRE would be something you'd want enabled all the time, hence having both. I think the main difficulty would be going through the existing ones and deciding which one each should be, though that can be deferred.

PR #404, opened 2024-05-23
Make additional use of `ParameterGroup`

mabruzzo:
### Pull request summary

Modify some existing parsing logic to make use of `ParameterGroup`.

### Detailed Description

The  `ParameterGroup` class was introduced back in #341 (the main body of that PR calls the class `ParameterAccessor` -- the name was changed based on feedback). It is a class whose sole-purpose is to provide access to parameters within a specific parameter-group (without specifying the full path to a parameter).

This functionality was duplicated in the pre-existing `Parameters` class. However, the `Parameters` class already provided a LOT of other functionality and this particular functionality is achieved by managing a bunch of internal state.

Part of the intention when introducing `ParameterGroup` was to eventually replace existing code that already made use of this functionality provided by the `Parameters` class so that it now makes use of the `ParametersGroup` class instead (in other words, the underlying logic is not changing). **This is what this PR does.**

The benefits are twofold:
- the code making use of this functionality is now a little more explicit
- it will allow us to remove this duplicated functionality from the `Parameters` class (this will actually be somewhat helpful for simplifying code in SMP-mode). 

**I plan to remove the duplicated functionality in a future, separate PR** since it will involve a number of modifications to existing unit-tests & it would be helpful to get the current changes into the main codebase. (Plus these changes have been kicking around on my computer for 3 months and I want to get them off my plate)

### DIFF
--- a/src/Cello/parameters_Config.cpp
+++ b/src/Cello/parameters_Config.cpp
@@ -327,12 +327,8 @@ void Config::read_adapt_ (Parameters * p) throw()
       (p->type(prefix+"schedule:var") != parameter_unknown);
 
     if (adapt_scheduled) {
-      p->group_set(0,"Adapt");
-      p->group_push(adapt_list[ia]);
-      p->group_push("schedule");
-      adapt_schedule_index[ia] = read_schedule_(p, prefix);
-      p->group_pop();
-      p->group_pop();
+      ParameterGroup p_group(*p, prefix + "schedule");
+      adapt_schedule_index[ia] = read_schedule_(p_group);
     } else {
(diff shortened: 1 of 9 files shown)


bwoshea:
@mabruzzo it seems that a bunch of tests are failing here.  It looks like it'll be easy to fix - could you do that when you get a second, and we'll then review this PR?

bwoshea (approved):
The changes are straightforward and I don't have any specific comments.  I will approve this PR, but I note that there is something wrong with CircleCI and I can't even SEE the tests, much less get them to run.  @mabruzzo if you can force the tests to run and they pass I think it's fine to merge this!

PR #406, opened 2024-06-03
Cleanup `EnzoInitialGrackleTest` and port parameter retrieval

mabruzzo:
### Pull request summary

I cleaned up the contents of `EnzoInitialGrackleTest` and ported the relevant parameter parsing

### Detailed Description

I cleaned up the following things within `EnzoInitialGrackleTest`:

- within the `EnzoInitialGrackleTest::enforce_block` method, I first removed calculations of temperature and pressure.
  - this has been unnecessary for years at this point (I think this originally may have been necessary before Andrew introduced the derived-field-output machinery).
  - this was also the only reason that the `EnzoComputePressure::compute_` existed, so I deleted that. I was also able to remove all other occurrences of the `grackle_field_data` struct from `EnzoComputePressure`
- I then removed all usage of the `grackle_field_data` struct from `EnzoInitialGrackleTest::enforce_block`. This let us remove the `ifdef CONFIG_USE_GRACKLE` statements from that method. I made a handful of other cleanup changes as well
- I also removed the `ifdef CONFIG_USE_GRACKLE` and warnings from `EnzoInitialGrackleTest::pup`. These warnings were entirely unnecessary (I think they are left over from before we had a coherent grackle-handling plan)

All of the above changes were done in order to achieve my goal of concentrating all grackle-related functionality into a tiny corner of the codebase and removing occurrences of `ifdef CONFIG_USE_GRACKLE` (once we are done, this will make it a lot easier to incrementally recompile Enzo-E with/without grackle).

Then I focused on porting the relevant parameter retrieval code out of `EnzoConfig` and into `EnzoInitialGrackleTest`'s constructor. This involved a little extra work since before this PR didn't actually store its own parameters (instead, it always retreived them from `EnzoConfig`)


bwoshea (approved):
This seems pretty straightforward to me.  Looks good!

PR #410, opened 2024-06-14
minor bugfix - Object's copy-assignment operation is currently wrong -- It doesn't return anything.

mabruzzo:
### Pull request summary

Fixing a minor bug in the ``Object`` class's copy-assignment operation. Currently it doesn't return a reference. I switched the implementation so that it now uses the default implementation.

### Detailed Description

While I was here, I modified the constructor and the copy-constructor to use the default implementation since there is no semantic difference (and it can sometimes produce faster code) 


gregbryan (approved):
Looks good to me -- any reason not to make the destructor the default as well?

PR #413, opened 2024-06-16
CMake Refactoring - Laying Foundations for (optional) Automatic Dependency Management

mabruzzo:
### Pull request summary

This PR introduces a number of minor changes related to the build system with the goal of insulating any external dependencies that are compiled as part of an Enzo-E-build from being passed arbitrary compiler flags that are passed to the source files that directly compose the core Enzo-E binary and the libraries in Enzo-E and Cello.

Currently, this doesn't make much practical difference.  But it lays the foundations for some forthcoming PRs -- in particular, I plan to add support for automatically installing Grackle (users will still be able to use their own external copy as well).

### Detailed Description

We have a number of build-system configuration options that customize the flags that are used by Enzo-E and Cello. Historically, we have specified these flags by 
- mutating the ``CMAKE_<LANG>_FLAGS`` variables (``<LANG>`` is replaced by ``C``, ``CXX``, or ``Fortran``) OR
- mutating the ``CMAKE_<LANG>_FLAGS_<CONFIG>`` variables (``<CONFIG>`` is refers to the build configuration -- ``Debug``, ``Release``, ``RelWithDebInfo`` or ``MinSizeRel``)

The challenge is that the flags encoded in these variables are indiscriminately applied language-wide to any files compiled as part of a build. 
- If we are automatically building some dependencies as part of the build, we probably don't want to apply all of these flags to those dependencies by default.[^1]
- Now, instead of modifying the above global variables, we use cmake's ``add_compile_options`` command.[^2] We use cmake's `$<CONFIG>` [generator expressions](https://cmake.org/cmake/help/v3.16/manual/cmake-generator-expressions.7.html#id2) to handle any flags that conditionally depend on build-configuration.

This PR consists of a few steps:
1. Factor out the specification of the minimal set of required fortran-compiler flags to properly handle the Enzo-E's fortran dialect.
    - Previously we handled this on a machine-by-machine basis -- it added to the boiler-plate of a machine configuration file. Basically the same logic was shared between all machines that used a given fortran-compiler
    - Now, the core build system have dedicated functionality that identifies the flags based on the compiler id. We no longer use the ``CMAKE_Fortran_FLAGS`` to specify these options
    - Factoring this out was not strictly necessary for this PR, but I think this is useful. The cmake-code was heavily inspired by some code I added to Grackle's build-system
2. Adjust of how exactly we apply floating-point optimization flags (we no longer mutate the ``COMPILE_<LANG>_FLAGS_<CONFIG>`` global variables) 
3. We moved away from using the ``CMAKE_<LANG>_FLAGS`` in the machine files to specify extraneous flags to making use of the new ``ENZOE_<LANG>_FLIST`` variables. It acts a lot like ``CMAKE_<LANG>_FLAGS`` but we only apply those flags while compiling the source files directly used by Enzo-E and Cello.
   - I wrote up some narrative documentation explaining that these variables are to be preferred by users and developers wishing to pass arbitrary flags to Enzo-E/Cello.
   - In the documentation, I tried to stress that these variables should only be used in a subset of cases (I want to make people aware that they exist, but ideally nobody should ever really need them outside of testing new flags or enabling extra warnings). Please let me know how well you think that is conveyed.
   - Maybe we should move the documentation to a subsection titled Advanced Usage?
   - I feel like the addition of these variables may border on being a little overkill... But I think we should probably be providing a way (without affecting the builds of any dependencies) to do this sort of thing.

4. I adjusted the location of the ``add_subdirectory("src/External)`` command in the main ``CMakeLists.txt`` file so that compilation of these files are not affected by the flags intended for Enzo-E and Cello. This wasn't strictly necessary (I don't think pngwriter is too sensitive to the flags in practice), but it was done for consistency.

5. I moved most of the dependency-finding logic to a new file called ``dependencies.cmake`` that gets included by the main ``CMakeLists.txt`` file. This is generally considered a best-practice for making the cmake-logic easier to follow.

This is a major change or addition (that needs 2 reviewers): probably not

### PR Checklist

Note that some of these check boxes may not apply to all pull requests.-->

[^1]: Here, I provide 2 motivating examples. First, consider the Fortran flags for specifying the Fortran dialect of Enzo-E's Fortran files -- we don't want to pass those flags to the Fortran in a dependency because they could be incompatible with the dependency's Fortran dialect. Second, if we are enabling warnings, we probably don't want to pass those warning-flags onto the dependency since that could produce warnings about the dependency's source code (that we don't care about while compiling Enzo-E)
[^2]: The semantics about which targets are affected by a given compile-option are much more intuitive when you specify it with ``add_compile_options`` than when you modify one of the global variables. For a handful of reasons, it would "better" to use the ``target_compile_options`` command instead, but that will involve a bunch of additional changes. With that said, the solution adopted in this PR is "good enough" for now (and acts as a good stepping-stone to that "better solution")

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -150,10 +150,10 @@ set(balancer_default "TreeLB" CACHE STRING "Charm++ load balancer to use by defa
 #TODO we should figure our reasonale defaults and/or provide instructions in the docs
 if (balance)
   foreach(BALANCER IN LISTS balancer_included)
-    string(APPEND Cello_TARGET_LINK_OPTIONS " -module ${BALANCER}")
+    list(APPEND Cello_TARGET_LINK_OPTIONS "SHELL:-module ${BALANCER}")
   endforeach()
   foreach(BALANCER IN LISTS balancer_default)
-    string(APPEND Cello_TARGET_LINK_OPTIONS " -balancer ${BALANCER}")
+    list(APPEND Cello_TARGET_LINK_OPTIONS "SHELL:-balancer ${BALANCER}")
   endforeach()
 endif()
 
--- a/doc/source/devel/debugging.rst
+++ b/doc/source/devel/debugging.rst
@@ -25,6 +25,13 @@ This will cause charm++ to open an xterm window for each node and within each wi
    You generally don't need to worry about having symbols for other libraries (like charm++).
 
 .. note::
+
+   If you are on a modern version macOS, you might encounter some weird errors complaining about the ``DISPLAY`` variable when you pass ``++debug`` or ``++debug-no-pause``.
+   One way to work around this issue is to install XQuartz from `the official website <https://www.xquartz.org/>`_ and download xterm through homebrew.
+   Note: you probably need to restart your machine after installing both of these.
+
+
+.. note::
    TODO: add instructions for using GDB on a remote cluster using an MPI-based build of charm++
 
    TODO: it may be useful to provide a link to documentation describing how to use GDB to investigate stack frames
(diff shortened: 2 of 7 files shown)


mabruzzo:
I think there was a weird mixup on my part.
- #414 seemed to have included the contents from this PR
- the code that I intended to merge as part of #414 seems to be gone.
- I remember doing some rebasing a while back and I must have broken things back then

Issue #426, opened 2025-07-10
[BUG] Double precision not always default configuration


clairekope:
Enzo-E seems to default to single precision, at least when compiling with `-DUSE_EXTERNAL_GRACKLE=ON`.

Full configuration:
`cmake -DCHARM_ROOT=~/research/charm/netlrts-linux-x86_64 -DEnzo-E_CONFIG=linux_gcc -DUSE_EXTERNAL_GRACKLE=ON -DGrackle_ROOT=~/local -B build/`

Error message:
```
CMake Warning at dependencies.cmake:54 (find_package):
  Found package configuration file:

    /home/claire/local/lib/cmake/Grackle/GrackleConfig.cmake

  but it set Grackle_FOUND to FALSE so package "Grackle" is considered to be
  NOT FOUND.  Reason given by package:

  Grackle was compiled with double precision, not single

Call Stack (most recent call first):
  CMakeLists.txt:59 (include)


CMake Error at dependencies.cmake:57 (message):
  Configured to use Grackle but Grackle library not found.

  Either disable grackle (e.g.  `-DUSE_GRACKLE=OFF`), start using automatic
  dependency management (e.g.  `-DUSE_EXTERNAL_GRACKLE=OFF), or provide path
  (e.g.  `-DGrackle_ROOT=/PATH/TO/GRACKLE/INSTALL/DIRECTORY`).
Call Stack (most recent call first):
  CMakeLists.txt:59 (include)
```

This error resolves if adding `-DUSE_DOUBLE_PREC=ON` to the configuration; however, the docs imply this should be on by default.

**Version information (please complete the following information):**
 - OS: Linux Mint
 - Charm++ version: 7
 - CMake version: 3.22.1
 - Grackle hash: 90d745160

gregbryan:
Thanks for pointing this out. It seems that it happens because the external library dependencies (line 59 of CMakeLists.txt)  are done before the option settings (in particular, the precision, which is set on line 66). @mabruzzo - is there any reason we can't just move the external library dependency check after (at least some) of the option setting part?

mabruzzo:
yeah, I think that's the exact problem! I think that would work

PR #428, opened 2025-08-13
Fixes issue #426, ensuring that Grackle is compiled with the correct precision (BUG FIX)

gregbryan:
### Pull request summary

Fixes issue #426, ensuring that Grackle is compiled with the correct precision

### Detailed Description

I moved the CMake precision setting code to after the external dependencies. It is possible this is not the optimal position, but it should at least resolve the issue.


mabruzzo (approved):
LGTM! I’m going to merge

